PR #9, opened 2014-03-05
overload connectionCheckUrl for use as base of plugin url

onejli:
implicitly use server hosting update site as connectionCheckUrl

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/HPI.java
+++ b/src/main/java/org/jvnet/hudson/update_center/HPI.java
@@ -57,8 +57,14 @@ public class HPI extends MavenArtifact {
     /**
      * Download a plugin via more intuitive URL. This also helps us track download counts.
      */
+    @Override
     public URL getURL() throws MalformedURLException {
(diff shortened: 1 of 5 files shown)


kohsuke:
Connection check is meant to test the reachability of the network, by hitting a server that's far more reliable than Jenkins infra. That's why it points to Google.

I believe your intent is to change where the artifacts get downloaded from, and if so, it should be added as a field to `MavenRepository` and `MavenArtifact` should get the value from there.

kohsuke:
And I forgot to mention, but my apologies to the delay.

daniel-beck:
@kohsuke A case can be made that `http://www.msftncsi.com/ncsi.txt` is a better default choice for users in countries with censored internet, see [JENKINS-23654](https://issues.jenkins-ci.org/browse/JENKINS-23654).

onejli:
@kohsuke @daniel-beck I'd proposed these changes because I'm operating an internal-only update site within a secure environment **without** access to the outside internet.  In cases like this, my thought had been that using the update site itself for the connection check url would be a relatively sane server-side check.  This is because at the time that I made the changes, Jenkins was relying on the client's browser to fetch the `update-center.json`, but was retrieving the plugins themselves via server-side operations (which introduces the possibility of being affected by different network ACLs for server and client).

PR #12, opened 2015-03-19
Ignore aws-lambda-jenkins-plugin and aws-lambda-jenkins

cast:
New plugin artifact id is aws-lambda to conform with artifact without -plugin guideline


orrc:
Ahh, this was renamed *twice*. That explains some of the weirdness I saw while working on #15.

daniel-beck:
@cast Could you please update this PR so it can be merged again?

daniel-beck:
Superseded by #18.

PR #13, opened 2015-04-13
Remove rtc

daniel-beck:
As stated on https://wiki.jenkins-ci.org/display/JENKINS/Rational+Team+Concert+Plugin for years:

> Deprecated since core 1.480.3.  Use Team Concert Plugin instead.


orrc:
FWIW, I did try to mark that plugin as deprecated via the wiki, but failed due to INFRA-275.

daniel-beck:
Can't hurt to have it excluded twice. Because we really mean it :wink:

PR #14, opened 2015-05-05
Add a lot of wiki overrides for plugins with incorrect/missing URLs.

orrc:
I changed the update centre code to be more strict about wiki URLs, and examined every single error that resulted...

Some plugins do have the correct URL in the POM, but haven't been released yet, a few are simple URL typos, but the majority specify no wiki URL in the POM.

Ideally somebody would also now make pull requests for most of the plugins on the list...

### DIFF
--- a/src/main/resources/wiki-overrides.properties
+++ b/src/main/resources/wiki-overrides.properties
@@ -28,6 +28,19 @@ script-scm=https://wiki.jenkins-ci.org/display/JENKINS/Script+SCM+Plugin
 # this plugin was forked from Checkstyle, but didn't modify the <url> tag...
 jshint-checkstyle=https://wiki.jenkins-ci.org/display/JENKINS/JSHint+Checkstyle+Plugin
 
+# Required until each of the Workflow plugins provide a URL in their pom.xml files
+workflow-aggregator=https://wiki.jenkins-ci.org/display/JENKINS/Workflow+Plugin
+workflow-api=https://wiki.jenkins-ci.org/display/JENKINS/Workflow+Plugin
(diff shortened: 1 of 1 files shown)


orrc:
Fixed a typo (`deployment-notification`) and removed one URL (`database-drizzle`) as the current codebase isn't good at handling `viewpage.action?pageId=123`-style URLs.

daniel-beck:
I have some doubts whether maintaining a central list of plugin/URL mappings is the correct approach going forward.

orrc:
I agree.

But it will be a while before all plugins have their POM updated *and* a new release made, so this is just an easy and instant fix for the ~50 plugins here that currently have *no* information at all in the plugin list aside from the artifact ID, e.g. "additional-identities-plugin", or display the plugin name but no wiki link or description.

Like it says at the very top of the file, ideally all entries in this file should be temporary.

olivergondza:
> Ideally somebody would also now make pull requests for most of the plugins on the list...

Adding wiki link where it is missing does not have to be reviewed by the maintainer, imho.

daniel-beck:
> Adding wiki link where it is missing does not have to be reviewed by the maintainer, imho.

As we discussed in the meeting yesterday, it's useless without a new release that includes the changed metadata, and maintainers may object to others directly committing to their plugin, so a PR would be best.

daniel-beck:
@orrc Do you think this should still be merged after the previous governance meeting?

orrc:
@daniel-beck Yes, this lets us grandfather in the existing releases with missing/bad URLs. 
As discussed, we should remove these overrides after we review in a few months.

I've fixed some UC bugs, and got in touch with some developers and updated this commit with more URLs.  If you could merge this now, that would be great.

PR #19, opened 2015-05-18
Add more plugin renames/deprecations.

orrc:


src/main/resources/artifact-ignores.properties
```diff
@@ -40,6 +40,20 @@ perfectomobile-jenkins    # renamed to perfectomobile
...
 sms-notification          # renamed to sms
 aws-lambda-jenkins-plugin # renamed to aws-lambda
-aws-lambda-jenkins        # renamed to aws-lambda
+aws-lambda-plugin         # renamed to aws-lambda
 aws-yum-paramater         # renamed to package-parameter
 aws-yum-parameter         # renamed to package-parameter
+hipchat-plugin            # renamed to hipchat
+hudson-logaction-plugin   # renamed to logaction-plugin
+icon-shim-plugin          # renamed to icon-shim
+mogotest                  # deprecated: mogotest service no longer exists
```
daniel-beck: There's no `mogotest` in http://updates.jenkins-ci.org/download/plugins/ ?

orrc: Ahh ok, that's because the UC generator only creates directories for plugins that are listed in the UC JSON.

The `mogotest` plugin does exist in Artifactory, and was used by a number of people:
http://stats.jenkins-ci.org/plugin-installation-trend/mogotest.stats.json

Because the [Mogotest wiki page](https://wiki.jenkins-ci.org/display/JENKINS/Mogotest+Plugin) is marked as deprecated, it's no longer in the UC.

*But* the UC only recognises the plugin as deprecated at the moment because it guesses what the `mogotest` wiki page URL is:

```
mogotest
** No wiki page specified.. picking one with similar name.
Using 'mogotest' for mogotest
```

Which is behaviour I've stopped in #20 — plugins now need to explicitly provide a wiki URL.
So I guess when that's merged, `mogotest` would be excluded anyway due to not having a valid wiki URL, but since it's never coming back, I though I may as well add it to the blacklist (and off the list of bad plugins without URLs).

daniel-beck: Makes sense.

PR #20, opened 2015-05-18
Exclude plugins without a valid wiki page from the Update Centre JSON files

orrc:
The criteria for inclusion are:
- Plugins must specify a Jenkins wiki URL in their `pom.xml`, via the `<url>` tag
- The wiki page must exist on the Jenkins wiki
- The wiki page must be a child of the "Plugins" page

Also accepted are:
- Wiki short URLs (e.g. `/x/GYCGAQ`)
- Wiki URLs with a numeric ID (e.g. `/pages/viewpage.action?pageId=60915753`)
- URLs which have legacy URL prefixes (e.g. non-HTTPS, wiki.hudson-ci.org)
- URLs ending with a trailing slash

Plugins not meeting these criteria will not be included in the Update Centre JSON file, or will have an empty `wiki` value in the release history JSON file.

As before, plugins may also be excluded via the `artifact-ignores.properties` file, or if a plugin does specify a valid wiki page, but that wiki page has the `plugin-deprecated` label.

While building the plugin list, more informational output is now printed to the console regarding why plugins are being excluded, or whether their wiki URL was rewritten to the canonical wiki URL format.

A summary of the total number of included, excluded and deprecated plugins is also shown at the end of building the Update Centre plugin list.

A number of test cases were added to test the basic URL resolution process, as well as some of the rarer types of URL found in currently-released plugins.

---

These changes are as agreed in the most recent community meeting (and broadly on the developers' mailing list).  This raises the quality bar a tiny bit to ensure plugins are at least minimally documented, and that users have a chance of seeing an excerpt in the Update Centre, and hopefully also source code links and JIRA info.

All current plugin releases which *do* have a wiki page, but *do not* list the URL in the POM have been grandfathered in via #14.

When running with `-no-experimental` (plus the overrides from #14, and the further exclusions from #19), this leaves **~~45~~ ~~38~~ 37 plugins** that would be excluded from the Update Centre, after this code is merged.  But the wiki overrides in #22 will provide a grace period for these plugins:

ID | POM URL
---|---
build-flow-toolbox-plugin | 
caliper-ci | https://code.google.com/p/caliper-ci/
chef | https://github.com/melezhik/chef-plugin
collabnet-automic-deploy | 
collabnet-uc4-deploy | 
ColumnsPlugin | http://wiki.jenkins-ci.org/display/JENKINS/ColumnPack+Plugin
cucumber | https://github.com/melezhik/cucumber-plugin
deploygate-plugin | 
[...]

PR #22, opened 2015-06-01
Add temporary overrides for plugins with no wiki page at all.

orrc:
Each of these plugins will, after #20 is merged, still be available in the UC, but the URL will point to the "[Plugin Documentation Missing](https://wiki.jenkins-ci.org/display/JENKINS/Plugin+Documentation+Missing)" wiki page, which tells users what the situation is, and warns that this plugin will be removed in future if no wiki page is added.


daniel-beck:
:+1:

PR #23, opened 2015-06-01
Add 1.609 UC

daniel-beck:
This needs to be done whenever a new 1.xxx.1 gets released, as http://updates.jenkins-ci.org/stable/latestCore.txt and in turn the information displayed on the web site depend on it.

### DIFF
--- a/site/generate.sh
+++ b/site/generate.sh
@@ -40,7 +40,7 @@ echo '<?php $rules = array( ' > "$RULE"
 #
 # Looking at statistics like http://stats.jenkins-ci.org/jenkins-stats/svg/201409-jenkins.svg,
 # I think three or four should be sufficient
-for v in 1.554 1.565 1.580 1.596; do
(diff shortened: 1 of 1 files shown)


site/generate.sh
```diff
@@ -40,7 +40,7 @@ echo '<?php $rules = array( ' > "$RULE"
 #
 # Looking at statistics like http://stats.jenkins-ci.org/jenkins-stats/svg/201409-jenkins.svg,
 # I think three or four should be sufficient
-for v in 1.554 1.565 1.580 1.596; do
+for v in 1.554 1.565 1.580 1.596 1.609; do
```
orrc: Should we also remove the oldest one at this point?
The comment above seems to suggest we should keep three or four in this list.

daniel-beck: @orrc Something to think about, but for now we need to get this new version in to fix the site and the stable UC.

orrc: @daniel-beck Sure, merge away.

PR #25, opened 2015-06-07
Add rally-bookkeeper

daniel-beck:
Removing rally-bookkeeper from the update center as requested by @mike-rogers.


mike-rogers:
This was intended to be a fork of the Rally Plugin that was "different enough" to warrant such a fork. After some discussion with core developers a new direction emerged and we decided instead to mark a breaking API change in the Rally Plugin and deprecate the older version. Please see https://github.com/jenkinsci/rally-plugin/pull/9 for more information.

On a personal note, I apologize for creating extra work for others involved.

oleg-nenashev:
:+1: . 
@daniel-beck  IMO it's better to use <code>@jenkinsci/core</code> instead of <code>@reviewbybees</code> if you need additional confirmations.

PR #26, opened 2015-07-06
Use the plugin name from each plugin's metadata, rather than wiki page titles

orrc:
Previously, if multiple plugins used the same wiki URL (e.g. workflow-*), then each of those plugins would use the same display name, leading to confusion when users saw multiple plugins with the same title in the Plugin Manager.

Now we use the name defined in the plugin metadata (e.g. POM `<name>`), and fall back to the wiki page title if that metadata is missing.  If the wiki page is missing or failed to be downloaded, we then fall back to using the artifact ID.


daniel-beck:
:+1:

PR #30, opened 2015-11-20
Deprecate the jenkow-plugin

batmat:


orrc:
Is this to be merged now, or are you waiting for feedback from the developer?  (I vaguely saw some related discussion in IRC)

batmat:
Good question, @rtyler and I chatted about it. He seems to think this would actually be enough to mark (hence, not merge this PR) the plugin wiki page with the `plugin-deprecated` label since you said previously this would also have the effect to hide it from the UC.

I'm personnally unclear, btw, what to do in general for plugin deprecation: should we recommend adding the wiki label ou submit a PR here? In which case?

In the current case, not sure the wiki label is going to be enough since this plugin is actually composed of 3 different plugins (like _workflow_ for example). So I don't know if the label on the wiki will actually hide the three.

WDYT?

orrc:
I think Tyler's [wiki page](https://wiki.jenkins-ci.org/display/JENKINS/Deleting+a+Plugin) is the right approach.

Using `plugin-deprecated` is enough for 99% of cases — as you mention, things like Workflow are different, so for rare cases like that we can still fall back to `artifact-overrides.properties`.

orrc:
Plugin has been removed via wiki tagging.

orrc:
Ack. Meant to command *and* close.

batmat:
Yup. Agreed.
Cheers
Le 26 nov. 2015 5:33 PM, "Christopher Orr" <notifications@github.com> a
écrit :

PR #32, opened 2015-11-26
Add option to skip release history creation

daniel-beck:
Disable history creation for all but latest update site

@orrc WDYT?


orrc:
:+1:

PR #38, opened 2016-01-13
Create a 1.642.x-specific site

jglick:
For example I have some changes prepared for `workflow-plugin` which require baselines after 1.625 (but before 1.642). I think this is the magic incantation.

@reviewbybees

### DIFF
--- a/site/generate.sh
+++ b/site/generate.sh
@@ -43,7 +43,7 @@ echo '<?php $rules = array( ' > "$RULE"
 
 # make sure the latest baseline version here is available as LTS and in the Maven index of the repo, 
 # otherwise it'll offer the weekly as update to a running LTS version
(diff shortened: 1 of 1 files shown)


jglick:
@olivergondza @daniel-beck seems like this is something that should just be done routinely as part of creation of a new LTS line, right?

daniel-beck:
@jglick Yes, it's on my to-do list for that.

Note that **this must not be merged before the next LTS.1 build is in the Maven repo cache**, otherwise it'll advertise the weekly build to all LTS users!

olivergondza:
So it should be part of release procedure for .1, right?

daniel-beck:
Not really, as there are at least several hours between releasing the bits and when this change should be published and the update sites generated.

PR #39, opened 2016-02-10
Add wiki override for the Play Framework plugin.

orrc:
Its wiki page was renamed without updating the POM and fell victim to [INFRA-301](https://issues.jenkins-ci.org/browse/INFRA-301).

This isn't really the plugin developer's fault, so we can add a temporary override.

Mailing list thread: https://groups.google.com/d/msg/jenkinsci-users/N3D0ceIor5M/PL61zc3IHAAJ


orrc:
@daniel-beck WDYT?

daniel-beck:
@orrc I agree.

PR #40, opened 2016-02-12
Update urls to use a secure connection.

goober:
Providing secure urls will enable jenkins installations that are installed behind a firewall that blocks outgoing requests on port 80 to use the update center. 

This is also relevant because the actual `update-center.json` by default in jenkins are requested through `https`.

### DIFF
--- a/site/generate.sh
+++ b/site/generate.sh
@@ -12,7 +12,7 @@ mvn -e clean install
 function generate() {
     java -jar target/update-center2-*-bin*/update-center2-*.jar \
       -id default \
-      -connectionCheckUrl http://www.google.com/ \
+      -connectionCheckUrl https://www.google.com/ \
       -key $SECRET/update-center.key \
       -certificate $SECRET/update-center.cert \
       "$@"
--- a/src/main/java/org/jvnet/hudson/update_center/HPI.java
+++ b/src/main/java/org/jvnet/hudson/update_center/HPI.java
@@ -58,7 +58,7 @@ public class HPI extends MavenArtifact {
      * Download a plugin via more intuitive URL. This also helps us track download counts.
      */
     public URL getURL() throws MalformedURLException {
-        return new URL("http://updates.jenkins-ci.org/download/plugins/"+artifact.artifactId+"/"+version+"/"+artifact.artifactId+".hpi");
+        return new URL("https://updates.jenkins-ci.org/download/plugins/"+artifact.artifactId+"/"+version+"/"+artifact.artifactId+".hpi");
     }
 
     /**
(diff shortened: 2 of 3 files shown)


orrc:
What kind of network do you have that blocks all traffic to port 80 but allows 443? :)

As per [INFRA-266](https://issues.jenkins-ci.org/browse/INFRA-266), this can't be implemented as the Jenkins mirror network does not fully support HTTPS.
Taking the Git plugin as an example:
```
$ curl -LI https://updates.jenkins-ci.org/download/plugins/git/2.4.2/git.hpi
HTTP/1.1 302 Found
Date: Fri, 12 Feb 2016 16:08:03 GMT
Server: Apache/2.2.14 (Ubuntu)
Location: http://mirrors.jenkins-ci.org/plugins/git/2.4.2/git.hpi
Vary: Accept-Encoding
Content-Type: text/html; charset=iso-8859-1

HTTP/1.1 302 Found
Date: Fri, 12 Feb 2016 16:08:04 GMT
Location: http://ftp.halifax.rwth-aachen.de/jenkins/plugins/git/2.4.2/git.hpi
```

goober:
Fair enough. Then I will follow the linked issue for a long term solution. However, what do you think about accepting the updated google URL to use a secure connection? In that case we could most probably open up the firewall for outgoing requests to your update site on port 80.

orrc:
Yeah, I look forward to having everything served over HTTPS!  We'll see how it goes...

There is a ticket about [changing](https://issues.jenkins-ci.org/browse/JENKINS-23654) the update site URL — it may not be necessary to check Google.

But I would be wary about changing the Google URL from port 80 to port 443, as I can imagine there are many installations that have access to port 80 without going through an HTTP proxy, but have access to port 443 blocked.  So we would be making one (failed) request to port 443 for the connection check, even although virtually all subsequent plugin download requests would be going via port 80.

In any case, I'm not sure that this would help you?

Even if you can pass the connectivity check by hitting Google on port 80 or 443, and you allow requests to updates.jenkins-ci.org on port 80 through your firewall, you'll still have problems because (like I showed above) you will likely be redirected to another hostname (mirrors.jenkins-ci.org), and from there you'll be redirected to a mirror whenever Jenkins tries to download the update site, tool installer data, or a plugin.

Depending on the status of the [mirror network](http://mirrors.jenkins-ci.org/status.html), and mirrors coming and going, you could end up being redirected to completely different hostnames over time, which would presumably not be accessible due to your firewall.

goober:
Thanks for the thorough feedback. I agree with you and will close this PR.
btw, have you ever considered bintray for plugin hosting? I saw that many of the jenkins plugins were available there.

PR #51, opened 2016-03-13
Hide unknown plugin flyway-runner

batmat:
See https://twitter.com/jenkins_release/status/708831455216992257


pskumar448:
@batmat @daniel-beck 
Now flyway-runner is forked into jenkinsci, could you please unblock the plugin so that it can visible in update center.

PR #58, opened 2016-04-19
Deprecate blackduck-installer

jrichardBD:
We are no longer supporting this plugin and it does not work with the current version of the update-sites-manager plugin, so we would like to remove it from the list of available plugins

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ target
 www
 .htaccess
 .idea
+/bin/
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@ target
 www
 .htaccess
 .idea
-/bin/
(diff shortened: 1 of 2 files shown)


batmat:
Why are there 3 commits instead of one?

jrichardBD:
They are 3 instead of one because I made the change and accidentally changed the .gitignore file, so I reverted the change and then made the change to the artifact-ignores.properties file again.

batmat:
In that case, in general you want to squash your commits or something. And in general, you also want to create a dedicated branch for each PR, that prevents you from adding unwanted commits to running PRs (didn't you have a similar _issue_ some time ago btw?)

jrichardBD:
OK, would you like to decline this PR and I will create a cleaner one? Or is this one line change alright?

batmat:
I don't have karma to handle that repository anyway. @rtyler did it.

Anyway, for the future, you may want to learn about rebasing and history rewriting in general with `Git`. It's very useful and helps keeping a readable history.

Cheers

jrichardBD:
I am sorry for the confusion and I will make sure any future requests are readable and clean.
Thank you!

PR #61, opened 2016-04-26
Active Directory 1.44 is broken

fbelzunc:
Active Directory plugin 1.44 is broken. See: [JENKINS-34426](https://issues.jenkins-ci.org/browse/JENKINS-34426)

I currently have the fix [here](https://github.com/jenkinsci/active-directory-plugin/pull/32) and have the confirmation from the person who opened the Jira issue that it is happening. I am waiting for review.

I think we should hide the broken version

@reviewbybees


daniel-beck:
Obsolete by now, I think.

PR #62, opened 2016-05-03
Escape HTML in plugin excerpts

daniel-beck:


src/main/java/org/jvnet/hudson/update_center/Plugin.java
```diff
@@ -285,6 +285,10 @@ public String getExcerptInHTML() {
             return null;
 
         String excerpt = m.group(1);
+
+        // escape malicious HTML
+        excerpt = excerpt.replace("<", "&lt;").replace(">", "&gt;");
```
rtyler: Do we have Apache commons here to use something like [escapeHtml4](http://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/StringEscapeUtils.html#escapeHtml4(java.lang.String)) ?

jglick: You forgot to escape `&`.

PR #63, opened 2016-05-03
Escape all the entities

daniel-beck:
@rtyler @jglick


batmat:
:+1:

rtyler:
Once @jglick signs off I think we can mergey merge

daniel-beck:
@rtyler Are we sure we need to wait? Worst case, I'll open a third PR for this line :laughing:

rtyler:
WHAT COULD POSSIBLY GO WRONG

PR #65, opened 2016-06-07
Adding flag to skip Wiki validation.

rafaelrezend:
I would like to create an _Update Center_ for a private repository. Even if I change the plugin repository and recompile the **backend-update-center2**, the execution still fails because it validates the existence of a wiki page (wiki.jenkins-ci.org) - for every plugin - as an acceptance condition. This pull request adds a flag to skip wiki validation.
I could live with this change on my side only (though I would prefer not to maintain my own fork), but I've noticed a trend in the Jenkins community to move the docs away from wikis into github itself. So, maybe in the future the community could benefit from this simple change (and hopefully other people willing to set up their own UC).

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -131,6 +131,9 @@ public class Main {
 
     @Option(name="-skip-release-history",usage="Skip generation of release history")
     public boolean skipReleaseHistory;
+    
+    @Option(name="-no-wiki",usage="Do not fetch plugin Wiki page")
+    public boolean noWiki;
 
     public Signer signer = new Signer();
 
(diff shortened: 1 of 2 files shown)


rafaelrezend:
Another possible implementation is to create an empty extension of ConfluencePluginList and use it whenever the "_-no-wiki_" flag is used. See [this commit](https://github.com/sjka/backend-update-center2/commit/4ad42fb4049dc02351a2cf15aa29e5133ddbe902) for details.

orrc:
Yeah, seems reasonable.

src/main/java/org/jvnet/hudson/update_center/Main.java
```diff
@@ -280,28 +284,38 @@ protected JSONObject buildPlugins(MavenRepository repository, LatestLinkBuilder
...
-                        } else {
-                            System.out.println(String.format("=> Excluding %s due to unknown/missing wiki URL: \"%s\"",
-                                    hpi.artifactId, givenUrl));
-                            missingWikiUrlCount++;
-                            continue;
+                System.out.println("Wiki page: " + givenUrl);
+                
+                if (givenUrl == null || givenUrl.isEmpty()) {
+                    System.out.println(String.format("=> No URL documentation provided for %s", hpi.artifactId));
+                    continue;
```
oleg-nenashev: I am not sure this is a correct behavior, because it suppresses `isVersionCappedRepository` and other logic below.

daniel-beck: I agree with Oleg, this change appears to be unnecessary and may have unintended side effects. #87 is attempting to solve a similar case differently.

daniel-beck:

PR #66, opened 2016-06-14
Don't cap core version unless it's necessary

daniel-beck:
Capping the version means that plugins violating the documentation rule are getting published. For LTS update sites, we accept this limitation (for now), but it's not necessary for the others.

Behavior should be no different than before, except that we're again filtering invalid plugin releases.

CC @kohsuke @rtyler who contributed to the code I'm removing, @batmat who reminded me of this issue, and @orrc who introduced the filter.


orrc:
Looks ok. Related: [INFRA-632](https://issues.jenkins-ci.org/browse/INFRA-632)

PR #67, opened 2016-06-15
[INFRA-800] Ignore jbehave hudson plugin

christ66:
This plugin was renamed to jbehave jenkins plugin.


rtyler:
@christ66 if you can update this commit with a ticket reference that has a bit more background, that would be useful

PR #68, opened 2016-06-20
Remove unknown release of build-timeout

daniel-beck:
Thanks to @oleg-nenashev

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -82,6 +82,7 @@ ez-templates-1.0.3
 ez-templates-1.0.4
 ez-templates-1.0.5
 jsontrigger               # not forked to jenkinsci; released without hosting request
+build-timeout-1.16-INTEL-1
 
 # broken releases, hpi/jpi does not exist or is not a zip file, resulting in maven errors
 accurev-0.6.28


oleg-nenashev:
👍 but actually thanks to @bsideup

bsideup:
Would be great to hear some postmortem about it. 

Was confused to found non-existing release (i.e. no tags in the code repository) :)

Also, exactly the reason why I created http://github.com/jucies/releases - to make it clear as hell where releases come from and associate them 1-to-1 with a code :D

oleg-nenashev:
I suppose @HedAurabesh was doing an internal release with fixes from the master and then messed up something in his Maven config. CC @ikedam

ikedam:
Can't we see who uploaded the version from access logs?

oleg-nenashev:
@ikedam See my comment above. It has been uploaded by @HedAurabesh

ikedam:
@oleg-nenashev Oh, you saw that from access logs. Thanks!

@HedAurabesh
Please have a separate environment to make private releases (e.g. another computer, or another  OS user, without credentials to the update center). It prevent you from uploading private releases to the update center accidentally.

oleg-nenashev:
@ikedam Just in case it happens again, it's possible to diagnose this issue using Artifactory. https://repo.jenkins-ci.org/webapp/#/artifacts/browse/tree/General/releases/org/jenkins-ci/plugins/build-timeout/1.16-INTEL-1

ikedam:
@oleg-nenashev It's really useful! Thank you.

PR #69, opened 2016-07-05
Adds an option to just verify that a plugin would be included.

abayer:
Given the argument "-verify-plugin someString", no files will actually
be written, core won't be processed, and only plugins containing
'someString' will be processed. This can be used to verify whether a
plugin release has a valid wiki page and meets any other criteria we
may add in the future.

Usage:

- mvn clean install
- java -jar target/update-center2-*-bin*/update-center2-*.jar \
    -id testing -verify-plugin github-branch-source

If any plugins matching the -verify-plugin string are found but don't
have valid wiki pages (currently that's the only thing we check
against, but if we add further criteria in the future, we'll check
those too), the script will notify the user and exit out with an exit
code of 1.

cc @rtyler @kohsuke @orrc @batmat

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -352,12 +352,20 @@ public class Main {
         System.out.println("Excluded " + deprecatedCount + " plugins marked as deprecated on the wiki.");
         System.out.println("Excluded " + missingWikiUrlCount + " plugins without a valid wiki URL.");
 
-        // If we are verifying a plugin string, check whether the missing wiki URL count is greater than 0.
-        if (verifyPluginStr != null && missingWikiUrlCount > 0) {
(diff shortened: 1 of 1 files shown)


rtyler:
I'm not the best person to review the code, but the idea I'm :+1: on!

abayer:
cc'd you as much for the idea as the code. =)

abayer:
...and the code could probably be more elegant. This was just the first approach that came to mind - it does work correctly (I verified by hacking `MavenRepositoryImpl` to force `github-branch-source` to use the problematic version 1.8, getting the correct, error-ridden result). My thinking is to have a parameterized job somewhere (doesn't even have to be on trusted.ci, could just as easily be on ci proper) that lets you specify the string and get a quick verification of whether there's a plugin matching/containing that string and that it would be included in the UC (or is deprecated - we don't care about deprecation in this context).

daniel-beck:
What's the point, as it requires the release to have happened already?

PR #73, opened 2016-07-29
[FIX INFRA-885] Add tool installers to versioned update sites

daniel-beck:
@rtyler @orrc @batmat


daniel-beck:
Seems I forgot `stable/` which gets special treatment.

site/generate.sh
```diff
@@ -54,10 +54,12 @@ for v in ${BASELINES[@]}; do
     # for mainline up to $v, which advertises the latest core
     generate -no-experimental -skip-release-history -www ./www2/$v -cap $v.999
     sanity-check ./www2/$v
+    ln -sf ../updates ./www2/$v/updates
```
rtyler: Because this is a relative symbolic link, I can't remember what will happen once it's copied into the docroot for Apache. :shipit:

PR #74, opened 2016-08-30
Add more broken releases

daniel-beck:
Attempting to fix https://groups.google.com/d/msg/jenkinsci-users/Xpe6oN5KDb0/UBHIfvUXAgAJ


rtyler:
@daniel-beck the thread only seems to reference one of the plugins blacklisted?

daniel-beck:
@rtyler You now approved the message explaining why it's three :-)

PR #79, opened 2016-09-16
Expose update-center.js at the top level

kohsuke:
Follow up to my previous PR that exposes `update-center.js`


kohsuke:
The file name is changing.

PR #81, opened 2016-09-16
Somebody pointed out that now we have .json and .js backward

kohsuke:
but we have to keep `update-center.json` for JSONP, so JSON version needs another name


rtyler:
Heh, somebody

PR #82, opened 2016-09-18
redirect update-center.actual.json to the current site

kohsuke:


daniel-beck:
🐜 Should escape the period AFAICT. Otherwise 👍

rtyler:
actually actual

PR #83, opened 2016-09-18
Fixed up regular expression match

kohsuke:
There's some strange use of character classes and '+' operator...


daniel-beck:
FTR `REQUEST_URI` excludes the query string, so should be fine.

👍

PR #87, opened 2016-10-06
Don't allow plugins without wiki pages into LTS UCs, if the plugin release is new

orrc:
If we're building a version-capped UC file, we now only include plugins without a wiki page if they were released before the wiki page rule came into force.

This would have fixed problems like the [Kubernetes plugins](https://issues.jenkins-ci.org/browse/HOSTING-113) showing up in the LTS UC, but not the latest UC (and therefore having no downloads on updates.jenkins.io), and also the gerrit-verify-status-reporter plugin from today.

So far in 2016, there have been releases of 8 plugins without valid wiki URLs, which are currently appearing in LTS UCs.  I've kept the existing behaviour for them, and set the cut-off time to October 2016 — this means that any plugins without a wiki page released from now onwards will not appear in the LTS UC.

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -293,9 +293,12 @@ public class Main {
                     if (actualUrl.isEmpty()) {
                         // When building older Update Centres (e.g. LTS releases), there will be a number of plugins which
                         // do not have wiki pages, even if the latest versions of those plugins *do* have wiki pages.
-                        // So here we keep the old behaviour: plugins without wiki pages are still kept.
+                        //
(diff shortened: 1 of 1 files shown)


daniel-beck:
What does this mean for new releases of these plugins?

* New release not shown, but old release remains
* New release not shown, no fallback, it disappears?

Should be accompanied by another reminder to the dev list, as well as PRs to the active plugins that don't have wiki pages.

daniel-beck:
@orrc Ping

daniel-beck:
What happens when a wiki page gets renamed, and a new version of the plugin released referring to the new URL, if there are plugin versions depending on an older core?

Will those just vanish from old LTS update sites, because they no longer refer to a valid URL?

daniel-beck:

PR #89, opened 2016-10-11
Prevent core version cap for main line update sites

daniel-beck:
The update center generator [will just use the plugin version cap if that's defined, but the capCore is not defined](https://github.com/jenkinsci/backend-update-center2/blob/27472026e75365a578b6d4903011b68e48d0b230/src/main/java/org/jvnet/hudson/update_center/Main.java#L167). So we need to set this to a very big value to get plugin versions cap without core versions cap.

Reference: https://groups.google.com/d/msg/jenkinsci-dev/xzFJ3nMhfa0/XEfTBK62BAAJ


daniel-beck:
@orrc WDYT?

bpedersen2:
Once deployed, https://issues.jenkins-ci.org/browse/JENKINS-36492 can be closed again.

PR #90, opened 2016-11-10
Suppress active-choices 1.5 and 1.5.0

kinow:
Hi, due to a regression in the plug-in, the Groovy script used for active-choices parameters is not being persisted after saving the job. This can cause users to lose their scripts after installing this version of the plug-in.

Thanks
Bruno


daniel-beck:
FTR @kinow is the author of that release:
https://github.com/jenkinsci/active-choices-plugin/commit/2fa6869224bcdd947939775e106e8cb322afbbcc

PR #92, opened 2016-12-10
Escape plugin title

daniel-beck:
Followup to #62 & #63.

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Plugin.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Plugin.java
@@ -316,6 +316,10 @@ public class Plugin {
         if (title == null) {
             title = artifactId;
         }
+
+        // escape malicious HTML
+        title = StringEscapeUtils.escapeHtml(title);
+
         return title;
     }
 


jglick:
Sounds reasonable. Did you find some with HTML titles, or this is just speculative?

daniel-beck:
@jglick 

I just compared the update center runs just before and after this PR was applied. No actual HTML tags were included anywhere, but a few plugins now have HTML entities in their titles:

      "title": "AbsInt a&sup3; Jenkins Plugin",
      "title": "AbsInt Astr&eacute;e Plugin for Jenkins",
      "title": "&quot;Anything Goes&quot; formatter",
      "title": "Dingding[&#38025;&#38025;] Plugin",

However it seems like I can (and should) undo this change, because…

screen shot

`j:out` and `st:out` are two very different tags, so [title](https://github.com/jenkinsci/jenkins/blob/fdba18915889d874c3bbfe21f726d0b0baee60ab/core/src/main/resources/hudson/PluginManager/table.jelly#L96) and [description](https://github.com/jenkinsci/jenkins/blob/fdba18915889d874c3bbfe21f726d0b0baee60ab/core/src/main/resources/hudson/PluginManager/table.jelly#L99) are handled very differently, something I did not notice. Only description/excerpt has the capacity for HTML.

PR #96, opened 2016-12-26
[INFRA-1022] Add support for warnings shown to Jenkins admins

daniel-beck:
List of warnings is initialized with recent plugin vulnerabilities

---

Companion PR to https://github.com/jenkinsci/jenkins/pull/2680 and https://github.com/jenkins-infra/backend-jenkins-plugin-info-plugin/pull/11

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -229,14 +229,10 @@ public class Main {
         return root;
     }
 
-    private JSONArray buildWarnings() {
-        try {
-            String warningsText = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream("warnings.json"));
-            JSONArray warnings = JSONArray.fromObject(warningsText);
-            return warnings;
-        } catch (IOException e) {
-            return new JSONArray();
-        }
+    private JSONArray buildWarnings() throws IOException {
+        String warningsText = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream("warnings.json"));
+        JSONArray warnings = JSONArray.fromObject(warningsText);
+        return warnings;
     }
 
     private static void writeToFile(String string, final File file) throws IOException {
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -27,6 +27,7 @@ import hudson.util.VersionNumber;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
 import org.kohsuke.args4j.ClassParser;
 import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
@@ -216,6 +217,7 @@ public class Main {
         JSONObject core = buildCore(repo, latest);
         if (core!=null)
             root.put("core", core);
+        root.put("warnings", buildWarnings());
         root.put("plugins", buildPlugins(repo, latest));
         root.put("id",id);
         if (connectionCheckUrl!=null)
@@ -225,6 +227,16 @@ public class Main {
             signer.sign(root);
 
         return root;
+    }
+
+    private JSONArray buildWarnings() {
+        try {
+            String warningsText = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream("warnings.json"));
+            JSONArray warnings = JSONArray.fromObject(warningsText);
+            return warnings;
+        } catch (IOException e) {
+            return new JSONArray();
+        }
     }
 
     private static void writeToFile(String string, final File file) throws IOException {
--- a/src/main/resources/warnings.json
+++ b/src/main/resources/warnings.json
@@ -28,7 +28,7 @@
   {
     "id": "SECURITY-258",
     "type": "plugin",
-    "name": "extra-columns",
+    "name": "script-security",
     "message": "Groovy sandbox protection incomplete",
     "url": "https://wiki.jenkins-ci.org/display/SECURITY/Jenkins+Security+Advisory+2016-04-11",
     "versions": [
--- a/src/main/resources/warnings.json
+++ b/src/main/resources/warnings.json
@@ -4,7 +4,7 @@
     "type": "plugin",
     "name": "google-login",
     "message": "Authentication bypass vulnerability",
-    "uri": "https://wiki.jenkins-ci.org/display/SECURITY/Jenkins+Security+Advisory+2015-10-12",
+    "url": "https://wiki.jenkins-ci.org/display/SECURITY/Jenkins+Security+Advisory+2015-10-12",
     "versions": [
       {
         "lastVersion": "1.1",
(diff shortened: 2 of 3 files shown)


daniel-beck:
@orrc @batmat WDYT?

Please only review this implementation here; and concept/data format/… in https://github.com/jenkinsci/jenkins/pull/2680

src/main/java/org/jvnet/hudson/update_center/Main.java
```diff
@@ -227,6 +229,16 @@ private JSONObject buildUpdateCenterJson(MavenRepository repo, LatestLinkBuilder
         return root;
     }
 
+    private JSONArray buildWarnings() {
+        try {
+            String warningsText = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream("warnings.json"));
+            JSONArray warnings = JSONArray.fromObject(warningsText);
```
oleg-nenashev: No check of the missing resource 🐛

src/main/java/org/jvnet/hudson/update_center/Main.java
```diff
@@ -227,6 +229,16 @@ private JSONObject buildUpdateCenterJson(MavenRepository repo, LatestLinkBuilder
...
+            return warnings;
+        } catch (IOException e) {
+            return new JSONArray();
```
oleg-nenashev: I would expect it to be logged somewhere

src/main/resources/warnings.json
```diff
@@ -0,0 +1,107 @@
...
+  {
+    "id": "SECURITY-136",
+    "type": "plugin",
+    "name": "extra-columns",
+    "message": "Stored XSS vulnerability",
+    "uri": "https://wiki.jenkins-ci.org/display/SECURITY/Jenkins+Security+Advisory+2016-04-11",
+    "versions": [
+      {
+        "lastVersion": "1.16",
+        "pattern": "1[.](\\d|1[0123456])(|[.-].*)"
```
oleg-nenashev: What will happen if somebody releases version 1.101? Just a corner-case, but we have plugins, which reach such numbers

src/main/java/org/jvnet/hudson/update_center/Main.java
daniel-beck: Uh… why would it ever not exist? Also, remember that this is a periodically run tool. If it doesn't exist, I'd hope it'd abort with error and fail the generation build.

src/main/resources/warnings.json
daniel-beck: It would not match the expression as it only expects two digits after the period, IOW it would behave as expected.

src/main/java/org/jvnet/hudson/update_center/Main.java
daniel-beck: Good point, as I wrote above generation should just fail in this case. I will modify it like that.

daniel-beck:
@oleg-nenashev 

> The companion PR is currently not tolerant against mistakes in the format of this warnings file. 

Irrelevant for this PR, no?

> I do not feel comfortable about releasing this feature without a file format static checker operating within the pull request builder.

Remember that file changes will only be done rarely and typically be accompanied by security advisories, when I'm already making sure the update site contain everything they're supposed to anyway. Therefore I don't see this as a big issue as e.g. changing artifact-ignores.properties to JSON would be.

daniel-beck:
@jglick points out in the core PR that it should be `url` rather than `uri` because it's expected to be a URL.

daniel-beck:
@oleg-nenashev 

Updated PR.

A PR builder can now trivially check the warnings file for correctness (mostly, wrong patterns will still be wrong) once implemented by running the test.

src/main/resources/warnings.json
```diff
@@ -0,0 +1,107 @@
...
+      }
+    ]
+  },
+  {
+    "id": "SECURITY-258",
+    "type": "plugin",
+    "name": "extra-columns",
```
daniel-beck: Copy & paste fail, needs to be `script-security`

michaelmccaskill:
Quick question: Is it possible to have the warnings attached to each plugin? Seems like it could be easily done.

daniel-beck:
@michaelmccaskill Not easily possible as we may publish warnings for plugins we're not publishing (to be shown to admins who already have that plugin installed). As we therefore need this top level element in any case, having another variant tied to the plugin element would just make the format and code handling them unnecessarily complex.

michaelmccaskill:
I see. Thanks for the consideration and explanation.

daniel-beck:
The core PR has been merged and KK started the release, so the format is sort of final.

@orrc @oleg-nenashev WDYT about the current state of the implementation?

daniel-beck:
@oleg-nenashev Not sure what you mean. This gets built before running, and failure to build (when the JSON is broken/incomplete) will abort the distribution.

Just needs a multibranch pipeline for PR builds, and we'll have the entire chain.

oleg-nenashev:
> Just needs a multibranch pipeline for PR builds, and we'll have the entire chain.

OK, I'm fine if there is a follow-up action item

src/test/java/org/jvnet/hudson/update_center/WarningsTest.java
```diff
@@ -0,0 +1,36 @@
...
+        String warningsText = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream("warnings.json"));
+        JSONArray warnings = JSONArray.fromObject(warningsText);
+
+        for (int i = 0 ; i < warnings.size() ; i++) {
+            JSONObject o = warnings.getJSONObject(i);
+            assertNonEmptyString(o.getString("id"));
+            assertNonEmptyString(o.getString("type"));
+            assertNonEmptyString(o.getString("name"));
+            assertNonEmptyString(o.getString("message"));
+            assertNonEmptyString(o.getString("url"));
```
oleg-nenashev: Nice2have: verify the advisory really exists

src/test/java/org/jvnet/hudson/update_center/WarningsTest.java
```diff
@@ -0,0 +1,36 @@
...
+public class WarningsTest {
+    public void testValidJsonFile() throws Exception {
+        String warningsText = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream("warnings.json"));
+        JSONArray warnings = JSONArray.fromObject(warningsText);
+
+        for (int i = 0 ; i < warnings.size() ; i++) {
+            JSONObject o = warnings.getJSONObject(i);
+            assertNonEmptyString(o.getString("id"));
+            assertNonEmptyString(o.getString("type"));
+            assertNonEmptyString(o.getString("name"));
```
oleg-nenashev: nice2have: verify the plugin really exists

daniel-beck:
@oleg-nenashev 

> OK, I'm fine if there is a follow-up action item

There is, sort of, just no ETA. Luckily this one should be fairly straightforward (unlike repo permissions), as no permissions are needed for a dry-run and it can therefore run on ci.jenkins.io.

src/test/java/org/jvnet/hudson/update_center/WarningsTest.java
```diff
@@ -0,0 +1,36 @@
...
+            assertNonEmptyString(o.getString("message"));
+            assertNonEmptyString(o.getString("url"));
+            JSONArray versions = o.getJSONArray("versions");
+            for (int j = 0 ; j < versions.size() ; j++) {
+                JSONObject version = versions.getJSONObject(j);
+                String pattern = version.getString("pattern");
+                assertNonEmptyString(pattern);
+                Pattern p = Pattern.compile(pattern);
```
oleg-nenashev: nice3have: go through the list of plugin versions and ensure the expression returns true and false at least once

PR #100, opened 2017-01-17
Ignore SCM API 2.0 related changes until JENKINS-41121 is addressed

rtyler:
The basic idea of this is that the upgrades are sufficiently dangerous and *one
direction* (i.e. no rollback except for restoring from a pre-upgrade backup of
JENKINS_HOME) that it's best for Jenkins users to hide these versions until this
is figured out and made more safe.

FYI @michaelneale @vivek  @i386

See https://jenkins.io/blog/2017/01/17/scm-api-2/ for the list of affected
plugins.

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -121,13 +121,13 @@ hello-world-scala
 
 # Suppress potentially harmful upgrades related to SCM API (JENKINS-41121)
 # Erring on the side of extreme caution until the fallout is resolved here
-scm-api:2.0.0
-scm-api:2.0.1
-branch-api:2.0.0
-github-branch-source:2.0.0
-cloudbees-bitbucket-branch-source:2.0.0
-github-organization-folder:1.6
-mercurial:1.58
-git:2.6.2
-git:3.0.2
-git:3.0.3
+scm-api-2.0.0
+scm-api-2.0.1
+branch-api-2.0.0
+github-branch-source-2.0.0
+cloudbees-bitbucket-branch-source-2.0.0
+github-organization-folder-1.6
+mercurial-1.58
+git-2.6.2
+git-3.0.2
+git-3.0.3
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -118,3 +118,16 @@ scm-api-2.0
 # Suppress Scala Hello World Plugin (HOSTING-235)
 hello-world-scala
 
+
+# Suppress potentially harmful upgrades related to SCM API (JENKINS-41121)
+# Erring on the side of extreme caution until the fallout is resolved here
+scm-api:2.0.0
+scm-api:2.0.1
+branch-api:2.0.0
+github-branch-source:2.0.0
+cloudbees-bitbucket-branch-source:2.0.0
+github-organization-folder:1.6
+mercurial:1.58
+git:2.6.2
+git:3.0.2
+git:3.0.3
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -124,6 +124,7 @@ hello-world-scala
 scm-api-2.0.0
 scm-api-2.0.1
 branch-api-2.0.0
+branch-api-2.0.1
 github-branch-source-2.0.0
 cloudbees-bitbucket-branch-source-2.0.0
 github-organization-folder-1.6
@@ -131,3 +132,5 @@ mercurial-1.58
 git-2.6.2
 git-3.0.2
 git-3.0.3
+workflow-multibranch-2.10
+pipeline-github-lib-1.0


michaelneale:
I tested it again with a small jenkins_home, and it did migrate fairly quickly. I am not sure if it works with large numbers of jobs or branches very well (which is why it appeared to not work).

michaelneale:
There is a problem with webhooks and branches with "/" in the name though that seems persistent...

rtyler:
@stephenc There is no "we cannot find them" in this case, they have been **found**, worse yet "in production."

As a project, I believe it to be reckless to leave known-unstable plugins exposed  it'is absolutely reckless (IMO) to leave known-unstable plugins which are as fundamental as these "in the wild." Considering these plugins are suggested to new users, and much of the Pipeline ecosystem rests upon them, I think this is a reasonable mitigation to avoid negatively impact to users.

New versions with fixes will automatically be made available since these are version-only ignores.


New update center JSON is regenerated in about 40 minutes. These regressions have now been in place for hours.

src/main/resources/artifact-ignores.properties
```diff
@@ -118,3 +118,16 @@ scm-api-2.0
 # Suppress Scala Hello World Plugin (HOSTING-235)
 hello-world-scala
 
+
+# Suppress potentially harmful upgrades related to SCM API (JENKINS-41121)
+# Erring on the side of extreme caution until the fallout is resolved here
+scm-api:2.0.0
```
orrc: While this syntax looks sensible, the only accepted syntax for this file is:
* `<artifact-id>`
* `<artifact-id>-<version>`

i.e. colons -> hyphens

daniel-beck:
@orrc I fixed the format.

orrc:
Looks fine; gave it a quick test, and `branch-api` was indeed rolled back to 1.11.1.

michaelneale:
The webhook problem has been reproduced and hopefully a fix incoming. 

Given that the remainder of the issues:

* the migration of dirs (risky, but seems ok)
* visual unpleasantness of URLs (URLs should redirect I think, so links should not be broken)
* a storm of reindexing

rtyler:
@daniel-beck is uncomfortable with the precedent being set of depublishing versions of plugins against its maintainers wishes. I understand this concern so I am going to close this pull request.


Separate from this, I plan on proposing additional rules and quality controls for plugins suggested as part of the Jenkins 2 Setup Wizard.

daniel-beck:
I'm in favor of this change in principle. It looks like the current release just isn't usable.

That said, the [Governance Document](https://wiki.jenkins-ci.org/display/JENKINS/Governance+Document#GovernanceDocument-Plugins) intentionally gives plugin maintainers a lot of decision-making authority for their plugin. Ignoring @stephenc's :-1: here sets a bad precedent:

* We're going to actively curate which plugin releases get distributed, and remove ones with compatibility concerns or major bugs from distribution.
* We are not opposed to overriding the maintainers' wishes in this case.

While the first one I could deal with, the second one I don't like at all.

So while @stephenc's :-1: stands, so does mine.

:-1:

src/main/resources/artifact-ignores.properties
```diff
@@ -118,3 +118,16 @@ scm-api-2.0
 # Suppress Scala Hello World Plugin (HOSTING-235)
 hello-world-scala
 
+
+# Suppress potentially harmful upgrades related to SCM API (JENKINS-41121)
+# Erring on the side of extreme caution until the fallout is resolved here
+scm-api-2.0.0
+scm-api-2.0.1
+branch-api-2.0.0
```
recampbell: Also needs branch-api-2.0.1

recampbell: @jglick says we should also block workflow-multibranch 2.10  should be blocked as well

abayer: And pipeline-github-lib 1.0.

HRMPW: Also Pipeline+GitHub+Library+Plugin 1.0

stephenc:
> Considering these plugins are suggested to new users

New users will not have issues (other than if they have ugly job names)

The only remaining "issues" are not "stability" but "devil and the deep blue sea" issues where there are two non-perfect choices and one must be picked.

Name mangling is the only option... unless we want to drop Windows as a supported platform for Jenkins servers...

URLs are preserved.

The "build storm" is actually Jenkins fixing the build result to assure that it has attached the correct build result due to bugs found in how it was determining what to build. The alternative is to actively suppress all builds between the last scan before upgrade and completion of the first scan after upgrade... and those commits will _never_ trigger a build. Triggering extra builds is annoying but the lesser evil than never triggering a build IMHO... but if people really feel strongly I am happy to merge and release the alternative option.

I am :-1: on merging this.

kohsuke:
I'm writing this as CTO of CloudBees.

This PR proposes a immediate mitigation to the significant concerns raised around the release of SCM API 2.0 and its impact on users.

SCM API 2.0 work championed by Stephen has happened as a part of the organized engineering activity by CloudBees. It's currently late in the night in the time zone Stephen is in, so the CloudBees engineering leadership met in the mean time and decided to withdraw Stephen's objection to temporarily block these plugins from the update center.

We take the impact to users very seriously, and this temporary withdrawal will buy time for our engineers to properly assess and evaluate the concerns raised, and address them properly if any additional work is necessary.  We think SCM API 2.0 represents a good work, and we just want to make sure it gets in the hands of users smoothly.

abayer:
+1

kohsuke:
And now as an acting infra guy as Tyler is travelling, I'm merging this PR.

kohsuke:
UC #11107 on trusted.ci.jenkins.io will bring this change to updates.jenkins-ci.org. ETA is 43mins.

kohsuke:
OSUOSL mirrors are verified to be serving new data, and one mirror that's lagging behind (xmission) was disabled temporarily to prevent it from serving stale data. When it's mirror program wakes up it should pick up the new metadata.

PR #106, opened 2017-02-02
Need to spin another workflow-multibranch

stephenc:
@reviewbybees URGENT @rtyler


daniel-beck:
This looks like it will result in plenty of unsatisfied dependencies…?

stephenc:
@daniel-beck we need to roll 2.12 and we want to hold it back from the UC until Blue Ocean is done

daniel-beck:
What's broken about this?

stephenc:
2.11 is failing blue ocean tests
2.11-beta-1 wasn't
So we need to revert / fix the change in 2.11-beta-1...2.11 which will require a new release

daniel-beck:
Future release for a bugfix needs preemptive block.

PR #109, opened 2017-02-03
Return SCM APi 2.0.x and friends to Update Center

stephenc:
go2

@reviewbybees

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -118,49 +118,5 @@ scm-api-2.0
 # Suppress Scala Hello World Plugin (HOSTING-235)
 hello-world-scala
 
-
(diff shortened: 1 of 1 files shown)


daniel-beck:
@stephenc So… can be merged now or what?

stephenc:
@daniel-beck https://github.com/jenkins-infra/jenkins.io/pull/580 needs to be merged and published first... but once that is public we are good to go on this one

rtyler:
@stephenc I disagree, the site is rebuilt very quickly, whereas Update Center information takes a couple of hours to propagate. if we publish the blog post(s) first then users won't be able to see anything for quite some time

stephenc:
@rtyler ok, I defer to your judgement. If you would rather merge this first we are 👍

PR #111, opened 2017-02-18
Ignores additions from wiki

daniel-beck:
Additions from `plugin-deprecated` label in an effort to start relying less on the wiki for building the update site.

https://wiki.jenkins-ci.org/labels/viewlabel.action?ids=46727211&key=JENKINS&startIndex=0

@orrc @rtyler @batmat


daniel-beck:
Oops, rebased onto current master and forgot to remove a few duplicate entries in the first comment, they are now in the second one, so https://github.com/jenkinsci/backend-update-center2/pull/111/commits/bab5405c9b222ed91efb1d6c1a179146bac81b24?w=1 isn't as empty as I wanted it to be.

orrc:
Sounds good. Individual commits look good to me.

PR #113, opened 2017-02-19
Remove support for the 'plugin-deprecated' label on the wiki

daniel-beck:
Only to be merged once #112 is, and we confirm no other plugins rely on wiki label deprecation.

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Plugin.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Plugin.java
@@ -254,22 +254,9 @@ public class Plugin {
         return labels;
     }
 
-    /** @return {@code true} if the plugin's wiki page has the "plugin-deprecated" label. */
-    public boolean isDeprecated() {
-        if (!labelsRead) readLabels();
-        return deprecated;
-    }
-
     private void readLabels() {
         if (page!=null)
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/update_center/Plugin.java
```diff
@@ -254,22 +254,9 @@ public String getScmHost() {
         return labels;
     }
 
-    /** @return {@code true} if the plugin's wiki page has the "plugin-deprecated" label. */
-    public boolean isDeprecated() {
```
oleg-nenashev: Is there a risk of impacting custom update centers using this class as API?
I see some usages of the codebase, but only in forks. So we unlikely care

rtyler: IMHO the Jenkins project doesn't support custom update centers (especially closed/proprietary ones) and that isn't something we should worry about too much here.

daniel-beck: They'll just need to adapt their code, either if a fork, or if using the update center as a dependency. If they set this up such that their stuff automatically breaks, they've done it wrong.

daniel-beck:

PR #114, opened 2017-03-02
Add baseline for new LTS line

olivergondza:
To be merged after 2.46.1 release. cc @daniel-beck


daniel-beck:
We should probably consider removing the update sites predating 2.0. It'll be a year (which is what we originally started with in dd88dc7131640203f5bd15f0b1353562eb2a237c -- and 2d67efa26858ea838e54de245ff1fd4a14be24fb rotated out the oldest one).

Then I took over, and just didn't bother doing that for reasons now unknown to me…

Still, :+1: for the actual change.

olivergondza:
Some of our instances still depend on 1.609 and we will depend on 1.651 for some time so I would better not to. Is it the pressure on the infrastructure that concerns you? I am fine dropping 1.5??, though.

daniel-beck:
> Is it the pressure on the infrastructure that concerns you?

Mostly annoying to have to wait forever for builds to show anything useful when it's time sensitive. But I'm working on inverting the order of update center builds, so the newest baseline starts first.

PR #115, opened 2017-03-03
Optimize execution

daniel-beck:
I noticed extreme amounts of traffic when building an update site locally, even with the minimal possible output (`-skip-release-history` set, `-www-download` or `-download` not set).

Sure, there's `-maxPlugins`, but I wanted the full output.

Turns out the update center build…
- downloads all plugin HPIs to compare to the version cap, even if that is undefined (i.e. `999.999`).
- downloads all plugin HPIs to find the two with the newest versions for further use.

Both are unnecessary, and optimizing them away result in a good reduction in traffic and disk space use in the local Maven repository if `-cap` isn't set.

Also remove a check nobody cares about related to when an artifact was created -- there's maintenance/stable branches, so that check seems invalid even if anyone cared.

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -272,7 +272,7 @@ public class Main {
         if (maxPlugins!=null)
             repo = new TruncatedMavenRepository(repo,maxPlugins);
         if (capPlugin !=null || getCapCore()!=null) {
-            VersionNumber vp = capPlugin==null ? ANY_VERSION : new VersionNumber(capPlugin);
+            VersionNumber vp = capPlugin==null ? null : new VersionNumber(capPlugin);
             VersionNumber vc = getCapCore()==null ? ANY_VERSION : new VersionNumber(getCapCore());
             repo = new VersionCappedMavenRepository(repo, vp, vc);
         }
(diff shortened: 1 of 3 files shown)


daniel-beck:
With the second commit, version-capped repositories are also optimized:

When determining the applicability of plugin releases based on their core dependency, we now only continue looking until we've found two: "current", and "previous". Once that's done, discard the rest and continue.

The order of repo nesting/filtering had to be changed to make this work; otherwise, one or more of the plugin versions retained could be alpha/beta, and filtered in a later step, leaving just one or none of the plugins. Changing the order prevents that (which makes more sense anyway, as alpha/beta is much cheaper to determine than core dependency).

This should also result in a slight speedup, and significant reduction of disk space, for production use, as no longer will all releases of all plugins need to be kept in the Maven cache, but only up to two per baseline (and typically much less).

Tested locally and saw no change in output (besides the weirdly TZ dependent UTC timestamps \*sigh\*).

daniel-beck:
Compared 'before' and 'after' output on updates.jenkins-ci.org (well, `wget --recursive --level 2 --reject '*.hpi,*.war' https://updates.jenkins-ci.org/`) to make sure this didn't break things, and it didn't. The only changes are index.html files (file timestamps in the few I checked) and two tool installers (one with random order of elements, the other got a new version between wget downloads).

PR #117, opened 2017-03-09
Make timestamps in output TZ independent

daniel-beck:
`ZipEntry#getTime()` uses the default TZ to convert the zip date to a unix epoch:

https://docs.oracle.com/javase/8/docs/api/java/util/zip/ZipEntry.html#getTime--

>If the entry is read from a ZIP file or ZIP file formatted input stream, this is the last modification time from the date and time fields of the zip file entry. The default TimeZone is used to convert the standard MS-DOS formatted date and time to the epoch time.

So to have a consistent time stamp in the output independent of the current system TZ, we should not set a time zone for the date formatter.


daniel-beck:
@orrc Assuming you're not in UTC, just run once with, once without, and compare to the output on updates.j.o. You'll be $utOffset off without this patch.

PR #118, opened 2017-03-10
Reference plugins.jenkins.io on release history

daniel-beck:
Will need a corresponding change downstream to link to the plugins site instead of the wiki: https://github.com/jenkins-infra/jenkins.io/pull/736

I added this as a new element as I don't know who might be using this.


rsandell:
The url for a plugin in core seems to [fallback to the wiki link](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/PluginWrapper.java#L366). Should we perhaps change that as well?

rsandell:
The [PlufinManager](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/PluginManager/table.jelly#L96) seems to use the wiki link as well.

daniel-beck:
@rsandell That's unrelated to any update center changes. That uses the URL defined in the pom.xml. So, 'Available' tab: plugins.jenkins.io. 'Installed' tab: Wherever the POM points.

PR #121, opened 2017-03-10
Add Jenkinsfile for PR builds

daniel-beck:
UNTESTED

Assumes https://github.com/jenkins-infra/backend-update-center2/pull/120 merged.

@rtyler WDYT? Or is the disk space requirement for the local Maven repo making this useless? Note that https://github.com/jenkins-infra/backend-update-center2/pull/115 could improve the situation.

Worst case we'll just build the generator, but not run it, in PRs.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,26 @@
+#!groovy
+
+properties([
+        buildDiscarder(logRotator(numToKeepStr: '10', artifactNumToKeepStr: '2')),
+])
+
+node('linux') {
+    stage('Prepare') {
+        deleteDir()
+        checkout scm
+    }
+
+    stage('Generate') {
+        withEnv([
+                "PATH+MVN=${tool 'mvn'}/bin",
+                "JAVA_HOME=${tool 'jdk8'}",
+                "PATH+JAVA=${tool 'jdk8'}/bin"
+        ]) {
+            sh 'mvn -e clean verify'
+        }
+    }
+
+    stage('Archive') {
+        archive 'target/surefire-reports/*-output.txt'
+    }
+}


Jenkinsfile
```diff
@@ -0,0 +1,30 @@
...
+
+    stage('Generate') {
+        withEnv([
+                "PATH+MVN=${tool 'mvn'}/bin",
+                "JAVA_HOME=${tool 'jdk8'}",
+                "PATH+GROOVY=${tool 'groovy'}/bin",
+                "PATH+JAVA=${tool 'jdk8'}/bin",
+                "SKIP_UPLOAD=true"
+        ]) {
+            sh './site/generate.sh'
```
oleg-nenashev: It uses -ex by default, right?

Jenkinsfile
```diff
@@ -0,0 +1,30 @@
+#!groovy
+
+properties([
+        buildDiscarder(logRotator(numToKeepStr: '10', artifactNumToKeepStr: '2')),
+])
+
+node('linux') {
```
rtyler: If this should really only run for PRs then I suggest wrapping this whole thing in a `if (env.CHANGE_ID) {`

daniel-beck: The current version now works on any host but doesn't actually do a lot -- just runs the tests ensuring the resources (primarily the plugins blacklist) make sense.

Jenkinsfile
```diff
@@ -0,0 +1,26 @@
+#!groovy
```
batmat: NIT: Pretty sure that in general Jesse does not like it ^. Maybe because, and I tend to agree, we should encourage people to think this is actual Groovy.

daniel-beck: > we should encourage people to think this is actual Groovy

Did you write the opposite of what you meant to write?

PR #124, opened 2017-03-13
[WIP] Optimize release history generation and timestamps

daniel-beck:
There are three related optimizations in this PR:

1. Only parse the plugin POM on demand rather than `Plugin` creation.
1. Use the artifact modification date rather than the `MANIFEST.MF` change time inside the `hpi`.
1. Do not include `title` or `wiki` in `release-history.json` entries older than a month

The **first change** should be fairly straightforward and comes with no change in behavior, it just enables the third one.

The **second change** will change almost all timestamps due to a time zone dependency I haven't figured out. Most release times change by 0-14 hours ± 5 minutes. Some, e.g. Surround SCM, change by weeks, something I don't understand and asked the maintainer at https://github.com/jenkinsci/Surround-SCM-plugin/commit/90a4e49f927e9444dceb4eaf54b897a2c414e7c7#commitcomment-21274381 -- basically, the release commit predates the upload date by 1-2 weeks here.

Only few current plugin releases (i.e. entries in update-center.json) are affected to that magnitude, and it's probably better to use a globally consistent reference date for plugins anyway.

One side effect if this is that the migration to Artifactory in 2010 will result in *many* older plugins having their release date set to April/May 2010 even if they were released earlier. Seems like an acceptable trade-off to me.

Finally, the **third change** is a straightforward optimization enabled by the first two: The update center generator needs to open the POM for the wiki URL and title, so skip that for old releases.

The only currently known consumer of this file is https://github.com/jenkins-infra/jenkins.io/blob/master/scripts/release.rss.groovy which ignores entries older than 30 days.

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
+++ b/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
@@ -38,8 +38,11 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.Calendar;
 import java.util.Date;
+import java.util.GregorianCalendar;
 import java.util.Locale;
+import java.util.TimeZone;
 import java.util.jar.Attributes;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
@@ -59,8 +62,6 @@ public class MavenArtifact {
     public final String version;
     private File hpi;
 
-    // lazily computed
-    private long timestamp;
     private Manifest manifest;
 
     public MavenArtifact(MavenRepository repository, ArtifactInfo artifact) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
```diff
@@ -165,27 +166,26 @@ public String getTimestampAsString() throws IOException {
...
-        Date tsDate;
         
-        try {
-            tsDate = bdf.parse(bdf.format(new Date(lastModified)));
-        } catch (ParseException pe) {
-            throw new IOException(pe.getMessage());
-        }
 
-        return tsDate;
+        Date lastModifiedDate = new Date(lastModified);
```
daniel-beck: Probably unnecessary.

src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
```diff
@@ -165,27 +166,26 @@ public String getTimestampAsString() throws IOException {
...
+        cal.set(Calendar.MINUTE, 0);
+        cal.set(Calendar.SECOND, 0);
+        cal.set(Calendar.MILLISECOND, 0);
+        return cal.getTime();
     }
     
     public static SimpleDateFormat getDateFormat() {
-        return new SimpleDateFormat("MMM dd, yyyy", Locale.US);
+        SimpleDateFormat sdf = new SimpleDateFormat("MMM dd, yyyy", Locale.US);
+        sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
```
daniel-beck: Time stamps appear to be one hour off from what's shown on Artifactory and I think this was intended to fix that, but didn't work out.

orrc:
Oops, thought this was still WIP. Anyway, 👍

daniel-beck:
@orrc Thanks. I forgot to remove the marker, it was never really WIP -- Just wanted to prevent premature merging. I checked the changelog contents and the last 30 days seem fine, some dates (and corresponding order) changed by up to a day, but otherwise same as before.

PR #126, opened 2017-03-13
[WEBSITE-332] Replace 'scm' host name by full browsable jenkinsci GitHub URL

daniel-beck:
**Do not merge** before people had some time to speak up and point us to consumers of the `scm` entry in the `update-center.json`, and we've adapted https://github.com/jenkins-infra/backend-jenkins-plugin-info-plugin to handle the new format.

Context:
https://issues.jenkins-ci.org/browse/WEBSITE-332
http://lists.jenkins-ci.org/pipermail/jenkins-infra/2017-March/001058.html

This tries to determine the SCM URL from the POM or parent POM's `scm/url` or `scm/developerConnection` elements. URLs pointing to old SVN repos are discarded, GitHub.com/jenkinsci URLs are rewritten to https:// (especially if SSH connection strings).

1100+ plugins have SCM URLs after this change (untested, some may be wrong), ~100 cannot be determined or only have known obsolete URLs, and another ~110 have URLs outside the Jenkins community (e.g. Bitbucket).

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Plugin.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Plugin.java
@@ -25,6 +25,8 @@ package org.jvnet.hudson.update_center;
 
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
+import org.apache.commons.httpclient.HttpClient;
+import org.apache.commons.httpclient.methods.GetMethod;
 import org.dom4j.Document;
 import org.dom4j.DocumentException;
 import org.dom4j.DocumentFactory;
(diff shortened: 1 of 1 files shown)


daniel-beck:
Amended to add fallback behavior and repository existence check.

- 16 plugins have SCM URLs that don't exist that are now caught by the HTTP check (well, 15 — one of them has an interpolation that's not supported).
- 117 additional plugins have repository URLs by implementing fallback behavior.
- 92 plugins have fallback URLs that don't exist.

So we went from 209 plugins without repository to 108 with this new change.

Based on these numbers, we could tolerate the 16 wrong URLs and instead decide to speed this up by only checking for fallback repository existence, not for every repo.

PR #127, opened 2017-03-14
Exclude renamed delphix-plugin.

orrc:
Duplicate plugin brought into the UC after the wiki page requirement was dropped; artifact ID was changed after version 1.0.0: https://github.com/jenkinsci/delphix-plugin/commit/d80a744d8f6e2d8edf3f9a3f2e5009bf34455ca0


daniel-beck:
https://github.com/jenkinsci/delphix-plugin/blob/master/build.gradle#L25 :disappointed:

PR #128, opened 2017-03-14
Release broken by JENKINS-42739

fbelzunc:
@daniel-beck


src/main/resources/artifact-ignores.properties
```diff
@@ -132,6 +132,7 @@ matrix-project-1.4.1      # this specific version is excluded for INFRA-250
 rebuild-1.2               # release troubles.. skip these
 rebuild-1.3               # ??
 ruby-runtime-0.13         # JENKINS-37353, JENKINS-37771 and JENKINS-38145
+active-directory-2.1	  # JENKINS-42739
```
orrc: Nitpicking: This section should remain in alphabetical order, and should use spaces rather than tabs/spaces before the comment.

PR #132, opened 2017-03-27
Ignore these until launch date

michaelneale:
Please review @vivek @i386 @rtyler 

This ignores 1.0.0/1 from UC temporarily as discussed - artifacts need to match those in https://github.com/jenkinsci/blueocean-plugin (please cross check vivek and james). 

If this is merged, there will have to be a follow on PR to remove them on the 5th April (US time/date).


vivek:
LGTM

michaelneale:
Closing this, as we have agreed on an alternate approach (and also avoids setting a precedent we don't want to set).

PR #136, opened 2017-04-27
SECURITY-445

daniel-beck:


daniel-beck:
@orrc Good find, thanks

> actually used by Jenkins

No, but the wiki and plugins site use this.

PR #138, opened 2017-05-06
Restore BuildGraph-View Plugin

pskumar448:
With BuildGraph-View-1.6 version dependency with Build-Flow-Plugin marked as optional.
Now with latest version there is no mandate dependency,

@daniel-beck 
Let me know whether this suffix to restore the plugin.

Thanks


daniel-beck:
@pskumar448 Sure!

FWIW It seems that the only necessary change was the one in the POM, the others seem unnecessary (without having tried it myself).

PR #140, opened 2017-05-16
Restore YouTrack plugin for versions >= 0.7.0 (SECURITY-464)

erikzielke:


daniel-beck:
@erikzielke Do you have a reference for the fix?

erikzielke:
What do you mean by reference? I have changed the youtrack plugin to use the Script Security Plugin for Groovy-code. So would like to have it enabled again

daniel-beck:
@erikzielke  
> I have changed the youtrack plugin to use the Script Security Plugin for Groovy-code.

Right, I'd like to review this. I don't think it would be great for either of us if it turned out to not actually be fixed after restoring distribution.

daniel-beck:
While I'm concerned about the continued existence of `commentText` in the code, I was unable to, as non-admin user, run arbitrary scripts in Youtrack Plugin 0.7.1, so restoring distribution.

Responsible disclosure guidelines in case there's an issue left: https://jenkins.io/security/

PR #142, opened 2017-05-18
More specific error messages when resolution fails

daniel-beck:
@kwhetstone may care, not sure.

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
+++ b/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
@@ -107,7 +107,7 @@ public class MavenArtifact {
         try {
             return repository.resolve(artifact,"pom", null);
         } catch (AbstractArtifactResolutionException e) {
-            throw (IOException)new IOException("Failed to resolve artifact "+artifact).initCause(e);
+            throw (IOException)new IOException("Failed to resolve artifact "+artifact+ " pom").initCause(e);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
```diff
@@ -107,15 +107,15 @@ public File resolvePOM() throws IOException {
         try {
             return repository.resolve(artifact,"pom", null);
         } catch (AbstractArtifactResolutionException e) {
-            throw (IOException)new IOException("Failed to resolve artifact "+artifact).initCause(e);
+            throw (IOException)new IOException("Failed to resolve artifact "+artifact+ " pom").initCause(e);
```
orrc: What on earth is going on with that cast? 😆

orrc:
More specific is more better 👍

src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
```diff
@@ -107,15 +107,15 @@ public File resolvePOM() throws IOException {
...
         }
     }
 
     public File resolveSources() throws IOException {
         try {
             return repository.resolve(artifact,"jar","sources");
         } catch (AbstractArtifactResolutionException e) {
-            throw (IOException)new IOException("Failed to resolve artifact "+artifact).initCause(e);
+            throw (IOException)new IOException("Failed to resolve artifact "+artifact+ " sources jar").initCause(e);
```
batmat: +1, same here BTW.

PR #146, opened 2017-05-30
Translation Assistance Plugin is practically unusable

recena:
I think this plugin should be blocked from Updates Centers. Why?

1. The main feature of this plugin is not longer usable. Announced in the [plugin website](https://wiki.jenkins-ci.org/display/JENKINS/Translation+Assistance+Plugin)
1. Has this plugin a maintainer?
1. This plugin **is forcing** to keep a specific HTML markup in all Jenkins views: `var f = document.getElementById('footer');`
```
<script>var footer = document.getElementById('l10n-footer');
    var f = document.getElementById('footer');
    f.insertBefore(footer, f.firstChild);
    footer.style.display = "block";

    var translation = {}; 
    translation.bundles = "";
    translation.detectedLocale = "es";

    function showTranslationDialog() {
      if(!translation.launchDialog)
        loadScript("/jenkins/plugin/translation/dialog.js");
      else
        translation.launchDialog();
      return false; 
    }
</script>
```

The proposal would be to block this plugin and avoid [this code](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/layout/layout.jelly#L275).

@reviewbybees

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -114,6 +114,7 @@ tepco-epuw                       # unused? -- https://wiki.jenkins-ci.org/displa
 TestExecuter                     # renamed to selected-tests-executor
 testflight                       # service discontinued -- https://wiki.jenkins-ci.org/display/JENKINS/Testflight+Plugin
 TestResultsAnalyzer              # renamed to test-results-analyzer
+translation                      # The server handling submitted translations has been shut down
 tusarnotifier                    # superseded by DTKit Plugin with automatic migration -- https://wiki.jenkins-ci.org/display/JENKINS/TusarNotifier
 uithemes                         # removal requested by tfennelly
 upload-artifacts-to-nexus        # renamed, see https://issues.jenkins-ci.org/browse/HOSTING-41?focusedCommentId=250383&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-250383


daniel-beck:
I'm not sure what to think of this change.

On the one hand, it's basically part of the API for `PageDecorator/footer.jelly` implementations given that it has an easily reusable ID, and is used in at least two plugins (`translation` and the unpublished `friend-of-jenkins` we directly send to donors per https://jenkins.io/donate/).

These plugins will just break with JS errors. Users who may still use Translation Assistance for local translations cannot do that anymore; the Friend of Jenkins footer entry will disappear, likely resulting in the need to update it and send the fixed version to anyone requesting it; and any (likely private) plugins using the same mechanism to add to the footer will break and need updates.

On the other hand, all of that is relatively low impact -- these aren't major Jenkins use cases. If the element is gone, relevant JS will break, but the UI won't look broken. I couldn't find any other plugins in GitHub or Artifactory that use this mechanism, although the Artifactory search was terrible and I'm sure left out results.

Ultimately, I have to ask, what would be gain from doing this? There's no upside to this PR in isolation, and even the downstream PR removes a single DOM element that is invisible. Nothing of any significance will change for the better.

So, while there's not really a significant downside here -- all changes are well within what we usually tolerate in levels of breaking changes, with the most worrisome issue I expect being upset past donors requesting updates* --, there's no upside that I can see here either.

Therefore I would tend to reject this change.

\* We have no way of knowing how commonly this plugin is installed -- basically a variant of https://issues.jenkins-ci.org/browse/INFRA-1198

jglick:
Where is this downstream PR? I guess you just mean, a hypothetical PR removing that one line? I agree that this would be a pointless incompatibility.

As to this plugin, why not just cut a release deleting the now-broken submission functionality, so that it is clear it is good only for local uploads?

daniel-beck:
> Where is this downstream PR? I guess you just mean, a hypothetical PR removing that one line?

Right, that seemed pretty much implied by

> 3. This plugin is forcing to keep a specific HTML markup in all Jenkins views: `var f = document.getElementById('footer');`

and

> The proposal would be to … avoid [this code](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/layout/layout.jelly#L275).

recena:
What I really want is to redesign the footer and this was the first blocker. I assume this plugin was a bundle plugin in some time and hence, we find these restrictions.

recena:
@jglick @daniel-beck IMHO, this message _The server handling submitted translations has been shut down. While this plugin is still useful for local translations, it's not possible to submit translations to the Jenkins project anymore._ should be enough to block this plugin. What it means, the own plugin does not make sense in this moment.

I don't see the value of keeping unusable plugins. A large number of plugins does not mean something positive but, the opposite.

recena:
@oleg-nenashev My goal is to remove the _coupling_ between plugin (practically unusable) and the main layout of Jenkins. It is conditioning all views.

Do we have any objective information about the usage (not installations) of this plugin?

oleg-nenashev:
I am pretty sure that my previous employer is still using it for local localization like small ad-hoc documentation adjustments. We used to run configuration-as-code for managing these files externally accross multiple Jenkins instances in our setup. Not sure if it is an "objective" information what you need, but I strongly oppose deleting functionality just to make small code improvements.

Regarding the UI side, see my comment above. The plugin can be updated for the admin-only use in the current state.

If you want to remove the particular mentioned piece of specific code, you know the "Jenkins way": make it extensible and move the code to the plugin.

recena:
oh man! How many time have you used this sentence "I am pretty sure that my previous employer is still using it ..."? The perfect sentence for avoiding / blocking new initiatives and improvements. This is being [a habit|https://github.com/jenkinsci/jenkins/pull/2852].

recena:
It is really frustrating to contribute on this OSS project.

oleg-nenashev:
@recena the discussion could be more constructive if respond to the feedback you get in a constructive way. Just a hint for the future

recena:
@oleg-nenashev _constructive_ here means: you have to do what the small group of committers want.

Please, provide objetive information not opinion about hypothetical scenarios. **Jenkins needs changes in the UI and it is being IMPOSSIBLE to do anything**.

My hint for you: If we something wants to be improved, something or someone has to change.

recena:
This OR should be understood inside of a collection of PR where the feedback provided is very similar:

- hypothetical cackward compatibility 
- you should do that in this way without providing technical reasons
- do not remove anything. Looks like more is better

I don't have words...

oleg-nenashev:
#donotfeedthetroll. I have provided a REAL scenario from a PRODUCTION instance. If it is not enough to consider this scenario, I am not sure how I can get your attention.

If you have any concerns about my behavior, feel free to follow the [Code of Conduct Reporting Process](https://jenkins.io/project/conduct/). If other project members think that I violate CoC or misuse my right to provide feedback, I will be ready to reconsider my approaches and the things I do in the community.

recena:
@oleg-nenashev It is fine. I don't have anymore for saying. It is my mistake for trying over and over to contribute to the project.

recena:
This has been discussed offline. I'm going to prepare a different path.

PR #145, opened 2017-05-25
Add option to filter for stable core releases

daniel-beck:
Prevents serving regular weeklies to LTS users like earlier today due to premature merge of https://github.com/jenkins-infra/backend-update-center2/pull/144

Preparation for https://issues.jenkins-ci.org/browse/INFRA-1193

@olivergondza  @rtyler

PR #144, opened 2017-05-25
Add baseline for new LTS line

olivergondza:

PR #147, opened 2017-06-02
Add 2.60 update site

daniel-beck:
Basically #144 again, but #145 now means that a premature merge of this (should me mergeable at any time) will not result in invalid LTS update sites offering the 2.60 weekly release.

The only effect until 2.60.1 is out will result in an additional layer for weekly releases, where 2.46 < X ≤ 2.60 will be offered fewer incompatible plugins (as we're still offering incompatible plugin updates rather than outdated plugins between LTS baselines) -- i.e. none that require 2.61+.

@olivergondza

### DIFF
--- a/site/generate.sh
+++ b/site/generate.sh
@@ -4,7 +4,7 @@
 # Used later for rsyncing updates
 UPDATES_SITE="updates.jenkins.io"
 RSYNC_USER="www-data"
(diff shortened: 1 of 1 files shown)


daniel-beck:
Of course I'm an idiot because I only now remember my plans to incorporate INFRA-1193 ☹️

daniel-beck:
Possibly superseded by #148.

PR #149, opened 2017-06-07
Fix advisory URL for JENKINS-44643.

orrc:


orrc:
Ugh, GitHub web UI is at fault for the newline at EOF.

PR #151, opened 2017-06-10
SECURITY-495 Restoring Ontrack plugin >= 2.31

dcoraboeuf:


daniel-beck:
@dcoraboeuf I reviewed and tested this plugin; a few remarks that may be useful to you:

* Security y/n takes precedence over sandbox y/n, so the checkbox still being present without security (and then being ignored AFAICT) seems weird.
* `JenkinsConnector` could probably use a few `@Whitelisted` methods.
* I'm not sure `configuring` with `withCurrentUser` during execution makes a lot of sense. Instead, this should be called on submission of the config form to associate with the user performing the configuration. This should also allow for approval without having to have a failed execution first. Obviously only works with inline script.

---

While the whitelisting of the entire DSL package seems a bit much, I haven't been able to break it, so I'm letting this back in.

(As usual, anyone discovering issues in any plugin should report that privately as described on https://jenkins.io/security/#reporting-vulnerabilities)

---

@dcoraboeuf Next step would be to update https://wiki.jenkins-ci.org/display/JENKINS/Script+Security+Support+in+Plugins accordingly. Thanks!

dcoraboeuf:
Thanks for the merge Daniel. Very much appreciated.

I'll take your remarks into account for the next iteration.

PR #152, opened 2017-06-12
The diskcheck plugin is literally the worst and should be eradicated.

abayer:
Let me explain a bit. 

https://github.com/jenkinsci/diskcheck-plugin/blob/master/src/main/java/org/jenkinsci/plugin/Diskcheck.java#L121

No, really. It calls `System.exit(0)`. I've talked to multiple users *today* (just today!) who have installed it and then poof, their masters go down. That...is not acceptable. So please, dear god, get rid of it. Forever.

cc @daniel-beck @rtyler

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -28,6 +28,7 @@ create-fingerprint-plugin        # renamed to create-fingerprint
 datadog-build-reporter           # renamed to datadog
 delphix-plugin                   # renamed to delphix
 description-column               # renamed to description-column-plugin
+diskcheck                        # This plugin seriously calls System.exit. It is terrible and should not exist. https://github.com/jenkinsci/diskcheck-plugin/blob/master/src/main/java/org/jenkinsci/plugin/Diskcheck.java#L121. Seriously.
 dockerhub                        # removal requested by teilo, superceded by dockerhub-notification
 drools                           # deprecated
 emotional-hudson                 # Superseded by Emotional Jenkins -- https://wiki.jenkins-ci.org/display/JENKINS/Emotional+Hudson+Plugin


rtyler:
If the plugin is up for adoption and actively causing issues for users, then I don't think it's a problem blacklisting it.

daniel-beck:
That's clearly not great. Has anyone pinged @mthak about it? It doesn't seem *actively* abandoned, https://wiki.jenkins-ci.org/display/JENKINS/DiskCheck+Plugin isn't labeled.

mthak:
Hi Daniel,

I agree it's not written correctly and is more like a hack than anything
else. But the idea behind this is worth it i think, We have been using it
at our company.

I have written a better version of this using groovy remote call so it does
not depend on Jenkins internal disk usage etc. so the disk usage data it
represents is much more accurate.

I can formalize it and make it more standard if that's ok.

-Manoj

oleg-nenashev:
If the plugin is being maintained, there is no reason to depublish it unless there is a serious security concern or deprecation of the integrated service. In thins case it would be great to just fix the issue.

@abayer Have you created an issue by the way? ;) I see no component in JIRA for it, but I can create it for @mthak if he is interested to continue maintaining the plugin

daniel-beck:
@mthak Specifically, the problem here is that the plugin will shut down Jenkins non-gracefully when the code path in the original comment above is hit. IOW, it seems dangerous to run the plugin in its current form.

So this thread is about removing the plugin from distribution.

As maintainer, what's your stance here? Do you plan to fix this? Should we just go ahead and stop distributing for now? What do you think?

daniel-beck:
@oleg-nenashev Well, there was precedent for suspending distribution of a clearly broken plugin in SCM API 2.0.

oleg-nenashev:
> Well, there was precedent for suspending distribution of a clearly broken plugin in SCM API 2.0.

IIRC it was requested by the maintainer. I agree this use-case deserves to exist in general, but I'd guess it is something requiring voting in the ML, not just the PR approval process

daniel-beck:
@oleg-nenashev You wrote:

>  is no reason to depublish it unless there is a serious security concern

All I'm saying is: we've done it before, in the case of a critical bug affecting pretty much everyone, while the plugin was being fixed.

Also, note that the maintainer is already in this thread.

abayer:
I'd strongly advocate for suspending distribution of this plugin until it's fixed to not call `System.exit`. I'm not saying we should give it the boot because of other shortcomings, just that taking down the master is unacceptable behavior that shouldn't be allowed out in the wild. The plugin is fixable, sure, but as it stands right now, it's just going to cause frustration and possibly real problems.

mthak:
It does not have system.exit in its current version, we fixed that bug longtime back, I have not looked at it in recent past, but yes it shall not cause any reboot.

I am happy to maintain/enhance the functionality if it's ok.

abayer:
@mthak Yes, it does - https://github.com/jenkinsci/diskcheck-plugin/blob/master/src/main/java/org/jenkinsci/plugin/Diskcheck.java#L121 - you may have an internal version that has it fixed, but what's available to Jenkins users still has the `System.exit`.

mthak:
I agree that i have not maintained this plugin properly and is in general
not the best implementation around here.

Give me a day to put the newer version  and a few bug fixes and we can take
it from there.

Best,
-Manoj

abayer:
@mthak All I'm concerned with is getting rid of the `System.exit` calls - anything else may or may not be good, but only the `System.exit` calls are so bad as to, in my mind, necessitate removal from the Update Center. So yes, please get that addressed ASAP.

mthak:
I agree Andrew,

I did not push the internal version to external github it's my fault.
Please give me me few hours and i will publish the source code. The version
0.26 that community is using does not have the system.exit.

abayer:
@mthak Yes, it does - check in https://repo.jenkins-ci.org/releases/org/jenkins-ci/plugins/diskcheck/0.26/diskcheck-0.26.jar. Also, you need to switch to using the Maven release plugin, not just `deploy`ing a release version.

mthak:
Hi Andrew,

I fixed a few of these hanging issues, please take a look at it once.
https://github.com/mthak/diskcheck-plugin/pull/5 I did not have the correct
list of people to add in PR.

Best,
Manoj

mthak:
Hi Andrew,

Did we change the artifactory server for hosting jenkins plugin?
http://maven.jenkins-ci.org:8081/content/repositories/releases/ is not
reachable. Can you please point me to the correct location

Best,
Manoj

mthak:
never mind, i figured it out.

-Manoj

daniel-beck:
@mthak @abayer any news here?

daniel-beck:
@mthak @abayer ping

mthak:
Hi Daniel,

I fixed the plugin with and updated a newer version 0.29. 

-Manoj

oleg-nenashev:
The plugin has been released in a very strange way: https://repo.jenkins-ci.org/webapp/#/artifacts/browse/tree/General/releases/org/jenkins-ci/plugins/diskcheck/0.29 . No sources, no javadoc. My recommendation would be to switch to Maven Release Plugin as it is configured in the default POM

mthak:
Sure Oleg,

I will do that , give me a day or two. 


Best,
Manoj

mthak:
I fixed a bug reported by user and released a new version 0.30 using maven
release plugin.

Please let me know if we see any inconsistency there.

-Manoj

oleg-nenashev:
Seems to be fine now, thanks

daniel-beck:
I assume this is obsolete.

PR #153, opened 2017-06-27
Remove mber plugin from distribution

daniel-beck:
This service appears to have been shut down:
https://wiki.jenkins.io/display/JENKINS/Mber+Plugin

CC @firepub-pfreeman @onefrankguy


daniel-beck:
No response in several days, but as the service appears shut down, should be safe to blacklist. Can always be restored.

PR #154, opened 2017-06-29
[JENKINS-45126] Remove Gradle plugin 1.27

daniel-beck:
https://issues.jenkins-ci.org/browse/JENKINS-45126

Something went wrong with the 1.27 release.

CC @wolfs


wolfs:
Thanks for the heads up. I'll release 1.27.1 today.

oleg-nenashev:
@daniel-beck FYI

wolfs:
@daniel-beck I just released version 1.27.1, which should fix the issue. Please go ahead and blacklist 1.27.

daniel-beck:
@wolfs Thanks for the update. I checked and it looks better.

Probably not really necessary since nobody is offered 1.27 anymore, but to prevent it from showing in the archive, and for whenever we add old version support…

PR #155, opened 2017-07-04
A bad deploy is bad

michaelneale:
cc @vivek @i386

What happened was a partial deploy - it failed during a release:perform in maven due to a full disk (no warning) - leaving it in a bad state. 1.1.4 will go out to replace it, but we don't know of another way to undo the damage

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -222,14 +222,11 @@ plumber
 
 # A bad deploy, and maven multi module and no staging means bad bad time
 blueocean-rest-impl-1.1.3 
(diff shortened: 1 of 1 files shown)


i386:
🐝 Looks good to me. I wish we had Nexus staging :(

daniel-beck:
`deployAtEnd`?

michaelneale:
@daniel-beck I have heard that deployAtEnd either doesnt' work, fails in a odd way, or doesn't work at all for multi module projects.  cc @stephenc, did I hear that from you?

michaelneale:
actually I need to try deployAtEnd - searched my email and found a note that it "may" work... will try with next beta, may save embarrasments like this.

PR #160, opened 2017-07-24
fixed SECURITY-479 https://github.com/jenkinsci/splunk-devops-plugin/3

fengxx:
fixed by https://github.com/jenkinsci/splunk-devops-plugin/pull/3
changes:
https://github.com/jenkinsci/splunk-devops-plugin/compare/jenkinsci:release/1.5.2...1.5.3

### DIFF
--- a/src/main/resources/warnings.json
+++ b/src/main/resources/warnings.json
@@ -673,7 +673,8 @@
     "url": "https://jenkins.io/security/advisory/2017-04-10/",
     "versions": [
       {
-        "pattern": ".*"
+        "lastVersion": "1.5.2",
+        "pattern": "1[.]([01234]|5[.][012])(|[.-].*)"
       }
     ]
   },


daniel-beck:
Will take a look ASAP. Once merged, don't forget to update https://wiki.jenkins-ci.org/display/JENKINS/Script+Security+Support+in+Plugins

fengxx:
Thanks for the prompt reply!

src/main/resources/warnings.json
```diff
@@ -673,7 +673,8 @@
     "url": "https://jenkins.io/security/advisory/2017-04-10/",
     "versions": [
       {
-        "pattern": ".*"
+        "lastVersion": "1.5.2",
+        "pattern": "1[.][012345](|[.-].*)"
```
daniel-beck: This makes no sense if the fix is in 1.5.3. What am I missing?

daniel-beck: To clarify, the regex is what matters to Jenkins -- `lastVersion` is just for documentation on the plugin site or wiki pages.

fengxx: I thought the check is <= lastVersion using the regex, it seems I am wrong.
how about I bounce the version and release 1.6?

daniel-beck: Jenkins itself matches what's installed to the regex. As this also matches 1.5.3, people with that release will still appear affected. This is a common case, you'll find several patterns dealing with this in the file.

A new version will not make this any more correct, people may be running 1.5.3 and we should not tell them they're unsafe.

daniel-beck:
@fengxx It seems GitHub does not notify about new commits…

PR #163, opened 2017-09-04
[JENKINS-46130] Remove delivery-pipeline-plugin-1.0.4

daniel-beck:
https://issues.jenkins-ci.org/browse/JENKINS-46130

Maintainers are not re-releasing a completely broken 1.0.4, so it's time this is blacklisted. I expected that to just be fixed and the world to move on, but more than three weeks later, it's not.

Apparent maintainers: @tommysdk @Crasher-ua

@oleg-nenashev @rtyler @orrc @batmat

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -190,6 +190,9 @@ hello-world-scala
 # Suppress sse-gateway 1.14  - it has unsatisfied dependency on pubsub-light 1.5
 sse-gateway-1.14
 
+# JENKINS-46130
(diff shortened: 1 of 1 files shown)


orrc:
Also pinging @patbos who is the default assignee in JIRA, and has been active in that issue.

tommysdk:
@daniel-beck @orrc Hi,

Sorry for the slow response as I've been away. I have therefore not looked into the issue either but will do as soon as is possible. Thank you for addressing this in the meanwhile.

oleg-nenashev:
Thanks @daniel-beck

PR #168, opened 2017-10-11
Limit supported releases from 10 to 7

daniel-beck:
I'm sick of waiting 40 minutes for this build to finish every time I release security updates.
Supporting ten release lines via the update sites is beyond ridiculous, nobody should be using Jenkins this old. Anything older than 2.46.2 is plain insane, so I'm being generous here: Four lines means an update once a year. That seems reasonable.

### DIFF
--- a/site/generate.sh
+++ b/site/generate.sh
@@ -10,7 +10,7 @@ chmod +x jq || { echo "Failed to make jq executable" >&2 ; exit 1; }
 
 set -o pipefail
 
-RELEASES=$( curl 'https://repo.jenkins-ci.org/api/search/versions?g=org.jenkins-ci.main&a=jenkins-core&repos=releases&v=?.*.1' | ./jq --raw-output '.results[].version' | head -n 4 | sort --version-sort ) || { echo "Failed to retrieve list of releases" >&2 ; exit 1 ; }
+RELEASES=$( curl 'https://repo.jenkins-ci.org/api/search/versions?g=org.jenkins-ci.main&a=jenkins-core&repos=releases&v=?.*.1' | ./jq --raw-output '.results[].version' | head -n 7 | sort --version-sort ) || { echo "Failed to retrieve list of releases" >&2 ; exit 1 ; }
 
 set +o pipefail
 
--- a/site/generate.sh
+++ b/site/generate.sh
@@ -10,7 +10,7 @@ chmod +x jq || { echo "Failed to make jq executable" >&2 ; exit 1; }
 
 set -o pipefail
 
-RELEASES=$( curl 'https://repo.jenkins-ci.org/api/search/versions?g=org.jenkins-ci.main&a=jenkins-core&repos=releases&v=?.*.1' | ./jq --raw-output '.results[].version' | head -n 10 | sort --version-sort ) || { echo "Failed to retrieve list of releases" >&2 ; exit 1 ; }
+RELEASES=$( curl 'https://repo.jenkins-ci.org/api/search/versions?g=org.jenkins-ci.main&a=jenkins-core&repos=releases&v=?.*.1' | ./jq --raw-output '.results[].version' | head -n 4 | sort --version-sort ) || { echo "Failed to retrieve list of releases" >&2 ; exit 1 ; }
 
 set +o pipefail
 


daniel-beck:
CC @olivergondza who might be impacted if he hasn't updated his instance since #148(ish).

daniel-beck:
@oleg-nenashev We've already had this discussion in https://github.com/jenkins-infra/backend-update-center2/pull/131 (with historical references that it used to be 4 baselines before I messed up).

The only reason we kept back to 1.609 then, and I started with 10 baselines once it became dynamic, is a request by @olivergondza – if he hasn't updated by now, he never will ☹️ 

By now, 
* I hate that we're currently, through our infra, *supporting* this use, [while not providing security updates](https://jenkins.io/security/advisories/). That's schizophrenic, and actively user hostile (some may only learn they're way behind if plugin updates all start saying "for newer Jenkins"). 
* Additionally, shipping security fixes on a tight schedule is much more difficult since the update center build now takes 30 minutes. Since a speedup in Artifactory Maven indexing (down to 10-15 minutes), it has become the slowest part of the entire process.

The data on plugin updates also indicates users on really old instances generally don't update anything, so they're not missing out by us not providing tailored update sites anymore. They'll still be able to update core, and install plugin updates, assuming their baselines are old enough.

olivergondza:
While I still have some instances on 1.609, updates there are not critical as you suggested. though my company still depends on 1.651 heavily (dozens of instances) BUT we are in the process of migrating them to 2.60 with update pace accelerating.

If I count correctly, that would be 7 update centers instead of 10. Would that be any better?

daniel-beck:
> If I count correctly, that would be 7 update centers instead of 10. Would that be any better?

Not really. I'd be willing to go to 5.

Also, note this isn't discussing which releases receive security updates, or anything like that. It's for which releases we're making it *very convenient* to *update specific plugins* through the UI. We're not shutting anyone's instances down.

olivergondza:
I do understand that, though we have internal Update Centers that are fed from the corresponding upstream ones. Let me discuss it internally.

daniel-beck:
To top it all off, security updates for plugins may not be available for these really old update sites, but given how we implement blacklisting (none at all, if fixed), unsafe versions continue being distributed in a fashion that allows one click installs, in releases pre-dating security warnings.

Since the warnings are also added for baselines older than 2.32 which do not ordinarily consume them, just set the update site URL to e.g. `https://updates.jenkins.io/stable-1.609/update-center.json` in a recent Jenkins and see what plugins are being distributed with vulnerabilities. (Spoiler alert: It's a train wreck.)

As reference, let's look at 2.19's update site (the one I'm proposing becomes the oldest supported baseline):

* `extreme-notification`, `app-detector`, `claim`, and `shared-objects` have active warnings for scripting issues affecting very rare configurations (admins who cannot run scripts) so continue being distributed.
* SAML Plugin (1k installs) suffers from a [low severity issue](https://jenkins.io/security/advisory/2017-08-08/) and the fix is only available to 2.60.1 or newer. Not great, but no big deal either.

And 1.609? It actively distributes *all the vulnerabilities* (or, well, it feels like it):

* analysis-core
* cas-plugin
* claim
* config-file-provider
* cucumber-reports
* datadog
* dependency-check-jenkins-plugin
* deploy
* distfork
* docker-commons
* dry
* email-ext
* extended-choice-parameter
* extreme-notification
* favorite
* git-client
* github-branch-source
* job-dsl
* mailer
* parameterized-trigger
* saml
* shared-objects
* sidebar-link
* swarm
* warnings
* workflow-cps
* youtrack

Even 1.651 suffers from high severity `workflow-cps` and `script-sandbox` vulnerabilities ([this one](https://jenkins.io/security/advisory/2017-08-07/#multiple-groovy-language-features-allowed-script-security-plugin-sandbox-bypass)).

----

Between 

1. the long build duration impacting the security release process, 
1. us not supporting baselines this old in any other way (filing a bug in 1.609? show it in 2.7x or go away!), 
1. this behavior *enabling* long-term unsafe use of obsolete releases, 
1. and actively distributing plugins with critical or high severity security vulnerabilities through non-archive channels,

it's time we stop doing this. There's no benefit at all from a project perspective, as this dilutes our message of "keep Jenkins updated" and considering only current LTS supported.

I propose we go with 1.651 for the next ~2 months (i.e. @olivergondza's proposal, dropping 1.609, 1.625, 1.642), and restrict it to five baselines once we release 2.8x.1 in early December (which will mean 2.32 and up, dropping 1.651, 2.7, and 2.19).

olivergondza:
> I propose we go with 1.651 for the next ~2 months (i.e. @olivergondza's proposal, dropping 1.609, 1.625, 1.642), and restrict it to five baselines once we release 2.8x.1 in early December (which will mean 2.32 and up, dropping 1.651, 2.7, and 2.19).

That would work. Thanks!

daniel-beck:
Consensus seems to be that we can go with 7 now, and 5 in ~2 months.

PR #171, opened 2017-11-16
remove loadrunner from uc

rpionke:
This closed-source plugin was accidentally uploaded and needs to be removed from UC.

Also see https://github.com/jenkins-infra/repository-permissions-updater/pull/506


daniel-beck:
I would prefer a bit more context for this, as every single source file is licensed as Apache 2.

rpionke:
We know that the source code is available under http://repo.jenkins-ci.org/public/de/tsystems/mms/apm/loadrunner/1.1/
Though the git repository for this plugin is not publicly available and we would like to stop distribution of this plugin and refer to the official HPE Plugin: https://plugins.jenkins.io/hp-application-automation-tools-plugin

daniel-beck:
> Though the git repository for this plugin is not publicly available

That's not a requirement for it to be open source. Its source code is public under an OSI-approved license, so there's nothing "closed source" about it.

That said, I was just confused by the PR comment above. We'll of course stop distributing this as requested.

I think with a total reported number of 27 installations we don't need further messaging to users here.

PR #173, opened 2017-12-05
Limit to five release lines

daniel-beck:
Followup to https://github.com/jenkins-infra/backend-update-center2/pull/168 based on the consensus reached there of going with five supported baselines starting with 2.89.x.

Today, we update sites for 2.73, 2.60, 2.46, 2.32, 2.19, 2.7, and 1.651.
Tomorrow, we will add 2.89, and rotate out 2.19, 2.7, and 1.651.

### DIFF
--- a/site/generate.sh
+++ b/site/generate.sh
@@ -10,7 +10,7 @@ chmod +x jq || { echo "Failed to make jq executable" >&2 ; exit 1; }
 
 set -o pipefail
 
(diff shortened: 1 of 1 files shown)


daniel-beck:
FTR, the current update site builds finish in ~18 minutes with 7 release lines. Should be down to 12-13 minutes with this change. Let's see how this goes.

PR #178, opened 2018-01-03
[SECURITY-294] Restore Artifact Deployer plugin distribution 

FranckG:
Restore Artifact Deployer plugin distribution for version 1.2 or higher


daniel-beck:
Groovy support removed in https://github.com/jenkinsci/artifactdeployer-plugin/pull/9

Pom also declares https://github.com/jenkinsci/artifactdeployer-plugin/blob/ee3935e3fa54a6bb6751f704e0a8684961b2946c/pom.xml#L101 so that's useful.

A few references to Groovy are still in the code, but one is a field/setter pair that appear unused, the rest are unused imports.

daniel-beck:
Version 0.40 exists but it did not include Groovy capabilities anymore.

PR #170, opened 2017-11-08
update claim plugin vulnerable version

Greybird:
I cannot close SECURITY-296 as I cannot see it myself. (I used JENKINS-43811 for that)
I edited https://wiki.jenkins.io/display/JENKINS/Script+Security+Support+in+Plugins with the information needed.
Please tell me if I need to do other things.

PR #181, opened 2018-01-24
Fix claim plugin vulnerable versions regex

Greybird:
#170 used a regex matching 2.1x as a vulnerable version, while 2.10 is the first fixed version.


src/main/resources/warnings.json
```diff
@@ -380,7 +380,7 @@
     "versions": [
       {
         "lastVersion": "2.9",
-        "pattern": "1[.].*(|[.-].*)|2[.][0-9].*(|[.-].*)"
+        "pattern": "1[.].*(|[.-].*)|2[.][0-9][^0-9](|[.-].*)"
```
daniel-beck: What's the purpose of `[^0-9]` here? Just leave it out.

Greybird:
hi @daniel-beck ,

Sorry for this, I took the wrong asumption that the regex was not matching the whole string, but only a subpart, and that I needed this [^0-9]. 
I checked how Jenkins was doing the match, and you are right, I do not need this.

Thanks for pointing me to this issue, I repushed a corrected version

Arnaud

PR #182, opened 2018-01-30
JENKINS-48357 - blacklist jira-plugin 2.5.1 

warden:


vivek:
LGTM, can we get it merged?

daniel-beck:
What has changed since this conversation?

	[2018-01-29 18:03:23]  <+oleg-nenashev>	danielbeck: How much justification do we need to blacklist the plugin from UC? JIRA Plugin 2.5.1 has serious issues (JENKINS-48357), but it was incompatible with other plugins in 2.5.0 as well...
	[2018-01-29 18:53:52]  <@danielbeck>	oleg-nenashev uh, so, we wouldn't actually fix anything if we blacklisted 2.5.1?
	[2018-01-29 19:25:15]  <+oleg-nenashev>	danielbeck: dunno :) Apparently all JIRA versions are unstable since the release of Apache HttpComponents Client 4.x API in other plugins. I am not sure what I can do excepting fixing the plugin
	[2018-01-29 19:50:53]  <@danielbeck>	oleg-nenashev would rather not blacklist something with unknown result... what's the point if it doesn't make things better?
	[2018-01-29 19:51:29]  <+oleg-nenashev>	danielbeck: OK, will see if I can get somebody working on the fix

michaelneale:
@oleg-nenashev is this not going to help at all even temporarily?

oleg-nenashev:
@michaelneale @daniel-beck 

Nothing really changed so far, but the maintainer decided it's better to blacklist 2.5.1

* 2.5.0 is consistent with ITSELF, not with OTHER plugins
* 2.5.1 is consistent with OTHER plugins, but not with ITSELF (it won't work in isolation)

Pick your poison

michaelneale:
@oleg-nenashev @daniel-beck ok in that case - is probably reasonable (JIRA plugin still is ill.. but if they think this helps, why not)

warden:
@daniel-beck I'm basing my blacklist choice on the fact that we're getting numerous bug reports from different sources (plugin conflicts) since the release of 2.5.1. Earlier we did not get any reports at all (well, apart from occasional class loading conflicts).

daniel-beck:
@warden @oleg-nenashev Sorry about that, did not realize @warden is the maintainer.

PR #184, opened 2018-02-01
[JENKINS-49323] Add cloudbees-jenkins-advisor to the blacklist

kwhetstone:
[JENKINS-49323](https://issues.jenkins-ci.org/browse/JENKINS-49323)

CloudBees is ending the CloudBees Jenkins Advisor service.  We need to remove the plugin from the update center.  Current users will be notified about the service ending.


abayer:
ping @daniel-beck @kohsuke @oleg-nenashev

PR #187, opened 2018-02-12
Simplify plugin display names

daniel-beck:
'Hudson', 'Jenkins', and 'Plugin' (or 'Plug-In') is redundant, so
remove them.

@slide @batmat @orrc WDYT?

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Plugin.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Plugin.java
@@ -399,7 +399,7 @@ public class Plugin {
         name = StringUtils.removeEndIgnoreCase(name, " Jenkins Plugin");
         name = StringUtils.removeEndIgnoreCase(name, " Plugin");
         name = StringUtils.removeEndIgnoreCase(name, " Plug-In");
-        name = name.replaceAll("[ -.!]+$", ""); // remove trailing punctuation e.g. for 'Acme Foo - Jenkins Plugin'
+        name = name.replaceAll("[- .!]+$", ""); // remove trailing punctuation e.g. for 'Acme Foo - Jenkins Plugin'
         return name;
     }
 
--- a/src/main/java/org/jvnet/hudson/update_center/Plugin.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Plugin.java
@@ -394,6 +394,7 @@ public class Plugin {
             title = StringUtils.removeEnd(title, " Plug-In");
             title = StringUtils.removeEnd(title, " Plug-in");
             title = StringUtils.removeEnd(title, " plug-in");
+            title = StringUtils.removeEnd(title, " Jenkins");
         }
         return title;
     }
--- a/src/main/java/org/jvnet/hudson/update_center/Plugin.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Plugin.java
@@ -386,15 +386,16 @@ public class Plugin {
         if (title == null) {
             title = artifactId;
         } else {
-            title = StringUtils.removeStart(title.trim(), "Jenkins ");
-            title = StringUtils.removeStart(title, "Hudson ");
-            title = StringUtils.removeEnd(title, " for Jenkins");
-            title = StringUtils.removeEnd(title, " Plugin");
-            title = StringUtils.removeEnd(title, " plugin");
-            title = StringUtils.removeEnd(title, " Plug-In");
-            title = StringUtils.removeEnd(title, " Plug-in");
-            title = StringUtils.removeEnd(title, " plug-in");
-            title = StringUtils.removeEnd(title, " Jenkins");
+            title = StringUtils.removeStart(title.trim(), "Jenkins");
+            title = StringUtils.removeStart(title.trim(), "Hudson");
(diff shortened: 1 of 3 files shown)


orrc:
For info: rules, some test cases, and sample output from when I proposed this a few years ago:
https://gist.github.com/orrc/7a6e2b0605306f6d215e

daniel-beck:
@orrc Thanks, forgot about that. Your rules are more comprehensive than these, will need to review them. I don't necessarily want to standardize spelling, just get rid of redundant content.

In the mean time I'll take your reference to "I proposed this years ago" to mean approval ;-)

src/main/java/org/jvnet/hudson/update_center/Plugin.java
```diff
@@ -379,11 +380,21 @@ public String getScmUrl() throws IOException {
...
+        } else {
+            title = StringUtils.removeStart(title.trim(), "Jenkins ");
+            title = StringUtils.removeStart(title, "Hudson ");
+            title = StringUtils.removeEnd(title, " for Jenkins");
+            title = StringUtils.removeEnd(title, " Plugin");
+            title = StringUtils.removeEnd(title, " plugin");
+            title = StringUtils.removeEnd(title, " Plug-In");
+            title = StringUtils.removeEnd(title, " Plug-in");
+            title = StringUtils.removeEnd(title, " plug-in");
+            title = StringUtils.removeEnd(title, " Jenkins");
```
Wadeck: Dunno exactly where each value come from, but in the list from @orrc there is also in the middle column (seems to be the one that is used here) : 
- ClearCase UCM **Plugin!** 
- Computer-queue**-plugin**
  - leiningen**-plugin**
- Cucumber json test reporting**.**  (dot could make the listing a bit weird)
  - Inedo BuildMaster plugin for Jenkins**.**
- Mail Commander **Plugin for Jenkins-ci**
- Maven Metadata **Plugin for Jenkins CI server**
- PVCS SCM **plugin for Hudson**
- Seapine Surround SCM **- Jenkins Plugin** (after removal of Jenkins Plugin, it remains just `Seapine Surround SCM -` )
- HudsonWindmill**Plugin**

daniel-beck: Current data is `jq -r '.plugins[] | .title'` of http://updates.jenkins-ci.org/update-center.actual.json

PVCS SCM and Seapine Surround SCM cleaned up their names enough so it looks reasonable. I couldn't find HudsonWindmillPlugin.

I'm fine with the rest looking weird and different, I doubt we'll see any of these a lot. In some cases, even their authors clearly didn't care what the display name would look like, so why should we?

https://plugins.jenkins.io/maven-metadata-plugin (~2700 installs)
https://plugins.jenkins.io/mailcommander (487)
https://plugins.jenkins.io/computer-queue-plugin (142)
https://plugins.jenkins.io/leiningen-plugin (191)
https://plugins.jenkins.io/inedo-buildmaster (15)
https://plugins.jenkins.io/inedo-proget (40)

Given that we won't be able to magically remove all redundant bits and stupid "formatting" like trailing periods that people come up with, I don't think rules as complex as those proposed by @orrc are useful.

The following would look weird under the current proposed rules:

> PTC Integrity CM - Jenkins Plugin
> Plugin Usage - Plugin

I plan to amend this to address the problem of trailing punctuation, but that's it.

Wadeck: Works perfectly for me 👍

orrc: > Given that we won't be able to magically remove all redundant bits and stupid "formatting" like trailing periods that people come up with, I don't think rules as complex as those proposed by @orrc are useful.

My code just took that into account with `.replaceAll("[ -.!]+$", "")`, and I don't see any other weird suffixes in the current titles.

> In the mean time I'll take your reference to "I proposed this years ago" to mean approval ;-)

Yeah, I definitely agree with the idea.  There was a dev list thread about this, but not sure why I ended up not opening a PR.

orrc:
I looked at the current plugin titles, and wrote a quick unit test:
https://gist.github.com/orrc/5bf7e3c3308e1d57b1eeea4820dfd4dc

Most things look fine, but there will be a few weird ones that don't behave as expected:
- ClearCase UCM Plugin!
- DaticalDB4Jenkins
- JenkinsLint Plugin
- Smart Jenkins

daniel-beck:
@orrc Thanks, I fixed it up some more.

daniel-beck:
I'm pretty happy about the state of things now.

Compared to the simple removal of `^(Jenkins|Hudson) ` and ` Plug-?in$`, the following plugin names are cleaned up further:

Before: https://gist.github.com/daniel-beck/38d8f690d569c2956db879503d70c639
After: https://gist.github.com/daniel-beck/b8e2a31a6c4cab82c645361d015a82e5

Not perfect, but good enough. As I wrote previously, we cannot standardize all possible insanity, and that's not the goal here.

I will merge this unless someone objects.

slide:
:+1: from me, it will be nice to have this cleaned up a bit more

PR #189, opened 2018-03-01
Stop distribution of foreman-node-sharing in favour of node-sharing

olivergondza:
https://github.com/jenkinsci/node-sharing-plugin/pull/65

CCing owner, @pjanouse.


daniel-beck:
@olivergondza Could you clarify whether you want to have this merged or are waiting for feedback from @pjanouse?

oleg-nenashev:
@daniel-beck ^

PR #192, opened 2018-03-19
public artifact sha1 + metadata side by side with plugins binary

ndeloof:
to be consumed by docker image "install-plugin.sh" and configuration-as-code for plugin installation.
see https://github.com/jenkinsci/configuration-as-code-plugin/issues/104

note: publishing hex version of artifact digest for easiest consumption from a shell script (`sha1sum`) vs base64 encoded one using in UC json.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,6 +175,11 @@
       <version>1.10.19</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>1.10</version>
+    </dependency>
   </dependencies>
 
   <repositories>
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -36,6 +36,7 @@ import org.kohsuke.args4j.Option;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.util.ArrayList;
@@ -324,6 +325,11 @@ public class Main {
                 if (download!=null) {
                     for (HPI v : hpi.artifacts.values()) {
                         stage(v, new File(download, "plugins/" + hpi.artifactId + "/" + v.version + "/" + hpi.artifactId + ".hpi"));
+
+                        final File manifest = new File(download, "plugins/" + hpi.artifactId + "/" + v.version + "/MANIFEST.MF");
+                        try (OutputStream out = new FileOutputStream(manifest)) {
+                            v.getManifest().write(out);
+                        }
                     }
                     if (!hpi.artifacts.isEmpty())
                         createLatestSymlink(hpi, plugin.latest);
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -35,10 +35,11 @@ import org.kohsuke.args4j.Option;
 
 import java.io.File;
 import java.io.FileOutputStream;
+import java.io.FileWriter;
 import java.io.IOException;
-import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
+import java.io.Writer;
 import java.util.ArrayList;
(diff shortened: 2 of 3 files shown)


oleg-nenashev:
SHA1 is considered to be not strong enough in 2018. Maybe we want to publish something stronger

daniel-beck:
This looks like it would massively impact the run time of the update site generation.

ndeloof:
@oleg-nenashev I'm fine with sha256 if you consider this would be a better candidate
@daniel-beck why ?

ndeloof:
@daniel-beck we could also get the artifact sha1 from nexus index to avoid computing from file

daniel-beck:
Not computing from files would be preferable, but we probably want both SHA-1 and SHA-256 given https://github.com/jenkins-infra/update-center2/pull/191. (Artifactory does not have SHA-512 checksums by default.)

Given that SHA-1 in the metadata is Base64, I wonder whether the file should mirror that choice for consistency. It's weird but at least it would be consistently weird.

ndeloof:
I also added MANIFEST as a side-artifact published by update center, as this would allow computation of dependency graph before actually starting to download hpi files.

daniel-beck:
Can't help it, all of this looks like a hack. It transforms the archives into something like an API.

ndeloof:
@daniel-beck as long as update center only expose "latest" we are constrained to such "hack".
I wish update site would have real APIs.

daniel-beck:
@ndeloof 

> as long as update center only expose "latest" we are constrained to such "hack".
I wish update site would have real APIs.

I suggest to provide additional JSON files with the metadata you need, or add it to the existing file (which will probably become too large). That's a proper solution. Dumping MANIFEST.MF to the downloads directory for programmatic consumption, effectively making it an API, is not.

ndeloof:
@daniel-beck so, converting MANIFEST.MF into a json format and uploading would be ok for you ?
reason I'd like to get the Manifest, and not (yet another) json metadata is because `hudson.PluginWrapper` do rely on such a Manifest, so allows to reuse existing codebase

daniel-beck:
This still has the problem that the directory structure of the archive downloads becomes an entry point for distributables.

Today, we can just implement minor changes to the update center generator and plugin site, and we can store plugin files at completely different URLs. With this change, not so much.

Am I overthinking this? @rtyler ?

rtyler:
@daniel-beck I'm not sure I can understand this discussion and weighing the pros and cons of it. I think you will need to jump on a Google Hangout and slowly explain it to me in small words so I can understand.

daniel-beck:
@rtyler Let's discuss this in next week's infra meeting.

ndeloof:
@daniel-beck @rtyler any decision made during infra meeting ?

the intent here is to provide an official way to get Metadata for a plugin based on it's version, based on hosting URL `/download/plugins/:shortname/:version/:shortname.hpi`. This URL is only a _de facto_ standard, actual link to binary plugin file is part of update center metadata. Such link is already used by docker image's [install-plugin](https://github.com/jenkinsci/docker/blob/master/install-plugins.sh) script.

The intent of this PR is to :

1. provide checksum for plugin binaries so a download can be checked. 
1. expose plugin's metadata so one can get them for arbitrary plugin version (UC metadata only do expose 'latest')

This second point assumes we are fine to make download URL format an actual standard for update center layout.

ndeloof:
a possible alternative to this:
update center do include metadata for plugins' `latest` + previous release version. Maybe we could introduce an additional `versions` list with links to dedicated metadata files. So any code that need it can just follow this link, and we don't enforce this download URL pattern in update center design.
wdyt @daniel-beck ?

ndeloof:
also read https://github.com/jenkinsci/jep/pull/59/files#diff-d37718df6e1dbf6f51cbde624d32e737R251 about the context

daniel-beck:
TBH I have no memory discussing this with Tyler… we may have just forgotten this. Next meeting would be next Monday.

ndeloof:
@daniel-beck any update on this topic (did you had this meeting with @rtyler ?)

ndeloof:
ok, so you preferred to implement this by yourself in 23801004896796a95b303ec8667f95f134fbab2e 
could have been pleasant to let me know :-\
I'll open another one for metadata as I still need them.

daniel-beck:
This seems orthogonal.

Please also note that I've revised it in #196 due to the bandwidth requirement.

PR #194, opened 2018-04-20
Update VersionCappedMavenRepository.java

biancl:
Fix the download war package version error.

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/VersionCappedMavenRepository.java
+++ b/src/main/java/org/jvnet/hudson/update_center/VersionCappedMavenRepository.java
@@ -45,7 +45,7 @@ public class VersionCappedMavenRepository extends MavenRepository {
 
     @Override
     public TreeMap<VersionNumber, HudsonWar> getHudsonWar() throws IOException, AbstractArtifactResolutionException {
-        return new TreeMap<VersionNumber, HudsonWar>(base.getHudsonWar().tailMap(capCore,true));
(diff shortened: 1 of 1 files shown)


daniel-beck:
@biancl Please provide more details justifying this change. It's unclear what you're referring to.

biancl:
ok,i commit it for when using the "cap" parameter or "capCore" parameter to specify the jenkins version,i think the update-center should not contains lower version wars or plugins than i specified ,just the opposite,the result download files contain lower version wars but not contain newer version wars.
should i commit an issue first ?

biancl:
okay,i commit it for when using the "cap" parameter or "capCore" parameter to specify the jenkins version,i think the update-center should not contains lower version wars or plugins than i specified ,just the opposite,the result download files contain lower version wars but not contain newer version wars.
should i commit an issue first ?

daniel-beck:
> i think the update-center should not contains lower version wars … than i specified

That's… exactly what the -cap and -capCore parameters are for. -capCore will make sure to include the newest core version that's not higher than the -capCore specified. -cap will ensure no plugin requires a newer release of Jenkins than specified.

It allows us to limit the LTS update sites to LTS core releases by specifying a -capCore of 2.107.999, and then it serves 2.107.1, 2.107.2, or whatever is the newest release there.

Similarly, the 2.89.x LTS update site serves no plugin releases that require a Jenkins newer than 2.89.x and would not be installable on that release line, since its -cap is 2.89.999.

biancl:
oh,oh,thank you very much.

PR #195, opened 2018-04-20
removing piketec-tpt plugin from blacklist after removing dependency …

PikeTec:
…to insecure and blacklisted copy-to-slave plugin

Hello,

the newest version of the piketec-tpt plugin has no needs for the copy-to-slave plugin so the dependency is removed (additional to adding other features).

Best Regards,
Joachim

Merge Commit of the feature branch to master (pom.xml): https://github.com/jenkinsci/piketec-tpt-plugin/commit/8b2c893814c108fe819163f9f167e15184f68716#diff-600376dffeb79835ede4a0b285078036

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -269,5 +269,4 @@ liquibase-runner
 perforce
 build-configurator  # depends on copy-to-slave
 lsf-cloud           # depends on copy-to-slave
-piketec-tpt         # depends on copy-to-slave
 reviewboard         # depends on perforce


PikeTec:
I am sorry about the big diff but when the blacklisting hit us we were in the middle of extendig our plugin and the copy-to-slave plugin was already not needed any more. So it was not really feasible to get a nice commit only for removing the dependency. I really hoped that it would be enough to see the missing dependecy in the the pom file.

Thank you for your hint to the wiki. We removed some fields that are no longer needed and prefered to have clean code insetad of transient fields. So the users will get a message about unreadable parts of the config but they have not to reconfigure the jobs.

PR #196, opened 2018-04-25
Don't record SHA-256, use base64 for SHA-512

daniel-beck:
Followup to #191 that hopefully is less bandwidth hungry:

- Skip SHA-256 which was mostly for convenience (compare to Artifactory checksums) anyway
- Use base64 encoding for SHA-512 to shave off bits at the cost of being weird

FYI @orrc (previous #191 reviewer), @rtyler

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
+++ b/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
@@ -25,11 +25,8 @@ package org.jvnet.hudson.update_center;
 
 import hudson.util.VersionNumber;
 import net.sf.json.JSONObject;
-import org.apache.commons.codec.BinaryDecoder;
 import org.apache.commons.codec.binary.Base64;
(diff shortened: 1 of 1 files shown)


daniel-beck:
@rtyler It's better, but still not great:

````
VARIANT							SIZE FOR PLAIN TEXT	SIZE GZIPPED	GROWTH
Update Center Master Build 6 (before any changes)	1.303.433 bytes		266.508 bytes
Update Center Master Build 7 (with SHA-256, SHA-512)	1.619.289 bytes		467.928 bytes	+24% / +76%
Pull Request 196 (SHA-512 in b64 only)			1.449.813 bytes		378.085 bytes	+11% / +42%
````

Numbers in all cases are for the artifical (read: typo'ed) Jenkinsfile baseline 2.79, so not comparable to real update-center.json.

daniel-beck:
Merging this for now, and adapting the core change.

PR #197, opened 2018-04-25
publish plugin metadata side-by-side with artifact

ndeloof:
In addition to all versions of plugins HPIs, update center to publish metadata as JSON (the same one consumed by `hudson.model.UpdateSite.Plugin`)

Primary use-case is for Configuration-as-Code plugin to be able to install specific version of a plugin with dependencies. 

also see https://github.com/jenkins-infra/update-center2/pull/192

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -319,6 +319,12 @@ public class Main {
                 pluginToDocumentationUrl.put(plugin.artifactId, plugin.getPluginUrl());
 
                 JSONObject json = plugin.toJSON();
+                final JSONObject versions = new JSONObject();
+                for (VersionNumber version : hpi.artifacts.keySet()) {
+                    versions.accumulate(version.toString(), "http://updates.jenkins-ci.org/download/plugins/" + hpi.artifactId + "/" + version + "/metadata.json");
+                }
+                json.accumulate("versions", versions);
+
                 System.out.println("=> " + json);
                 plugins.put(plugin.artifactId, json);
                 latest.add(plugin.artifactId+".hpi", plugin.latest.getURL().getPath());
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -35,9 +35,11 @@ import org.kohsuke.args4j.Option;
 
 import java.io.File;
 import java.io.FileOutputStream;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
+import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


ndeloof:
@daniel-beck @rtyler

daniel-beck:
Thinking about this from an API consumer standpoint, this still requires consumers to know

1. the disk layout
2. the versions of each plugin

The latter is easy for CasC specifically, where users are expected to provide version numbers as configuration (and you just tell them to figure this out).

These issues still make this weird as an API. Ideally there would be some way to get from a known location to the specific versions without cementing the disk layout we currently have, forever.

ndeloof:
@daniel-beck I tried to address this concern with 19164f4
Open questions : why does the update center include all plugins and version if you consider the only "API" entrypoint is the json metadata, which only reference latest version (+ "previous" for some weird reason) ? I assume everybody consider the UC file layout as "API" based on this lack of metadata, and the fact it never changed within one decade of Jenkins history.

ndeloof:
sample (additional) output in uc metadata :
```json
"plugins": {
    "absint-a3": {
...
      "versions": {
        "1.0.0": "http://updates.jenkins-ci.org/download/plugins/absint-a3/1.0.0/metadata.json",
        "1.0.1": "http://updates.jenkins-ci.org/download/plugins/absint-a3/1.0.1/metadata.json",
        "1.0.2": "http://updates.jenkins-ci.org/download/plugins/absint-a3/1.0.2/metadata.json",
        "1.0.3": "http://updates.jenkins-ci.org/download/plugins/absint-a3/1.0.3/metadata.json",
        "1.1.0": "http://updates.jenkins-ci.org/download/plugins/absint-a3/1.1.0/metadata.json"
      }
```

Sounds a reasonable increase in this (already huge) metadata file.
A subsequent improvement in jenkins-core _could_ rely on this to offer version selection when installing a plugin in PluginManager.

daniel-beck:
We can't really increase the size of `update-center.json`, as I learned in #191 (and what motivated #196).

ndeloof:
So what ? Isn't it time to just kill this beast ?

ndeloof:
If this JSON file is becoming way too big, why not just admin the filesystem layout is a natural API ?
We can also have JSON metadata per plugin with list of versions
This "one-does-all" metadata file should only include the strict minimum required for Jenkins' PluginManager to offer the installation UI.

daniel-beck:
@ndeloof 

@rtyler [recently pointed me to a paper of package manager security issues](https://twitter.com/agentdero/status/989164631993221120), and this approach seems susceptible to multiple potential problems outlined in there.

Now, granted, the fact that today there's no sane way to ensure what you download is what you want (unless it's the latest version offered) is an even bigger problem.

I will try to prototype a solution that would solve your problem and address my concerns.

daniel-beck:
Similar to #198, I expect this to never finish update center runs, when not using version capped repos, as that is where I implemented filtering for the newest two releases (and that is necessary for getting all versions).

daniel-beck:
Superseded by #206.

PR #198, opened 2018-04-27
Add plugins versions index

daniel-beck:
Work in progress. Untested. I may rewrite history.


daniel-beck:
@ndeloof WIP, but still, WDYT? Would this work for you?

ndeloof:
sounds a reasonable approach, this would partially address my needs (I still need a full plugin metadata json file so I can build a specific `hudson.model.UpdateSite.Plugin`). Maybe this file could offer a link to plugin metadatas for a specific version ?

notes: why have a top level `plugins` element in this json file, and not just a plain json array ?

daniel-beck:
@ndeloof 

> still need a full plugin metadata json file so I can build a specific

Ah, okay. I specifically went for the reduced content because I thought it wasn't needed. Can probably add more stuff around it to enable it.

> not just a plain json array

I *could* claim future extensibility, but really just 'cause I'm dumb and copied this together blindly.

Future extensibility however is why I'd rather not change it.

ndeloof:
> future extensibility 

what would you expect a `plugin-versions.json` file to contain but plugins informations ? Future metadata could just be exposed in another file if required. Not a blocker anyway if you prefer this approach.

daniel-beck:
Currently limited to 2 versions per plugin, probably because of an earlier performance improvement I cannot find right now. Will need to dig a bit.

daniel-beck:
> to contain but plugins informations ? Future metadata could just be exposed in another file if required

Which sort of contradicts your requirement for full (and mostly redundant) metadata per version. Just grab the rest from update-center.json? Would probably be preferable in case of name/description/maintainer/… changes anyway to always use latest.

ndeloof:
You don't get my point : I don't understand the need for a top level `plugin` element in a file used to list ... plugins. Should just be a json array. And if we later want to expose more metadata about anything else, another json file can be created. I don't consider redundancy to be an issue here. I don't expect to get all plugin metadata served by this file, keep it small enough as a list of plugins versions (as the name claims) with links to plugin metadata json file for each version. 

Anyway, do as you feel comfortable as long as I can get metadata for each and every plugin hosted in Update Center, which is not the case today (and make me wonder why this update center do host those files, if the directory structure is not considered "stable")

daniel-beck:
> You don't get my point : I don't understand the need for a top level plugin element in a file used to list ... plugins. Should just be a json array.

I don't feel strongly, can do that. Just seems like it sacrifices flexibility and extensibility to save about 5 bytes.

> I don't expect to get all plugin metadata served by this file, keep it small enough as a list of plugins versions (as the name claims) with links to plugin metadata json file for each version.

As I wrote before, this presents problems related to verification of the metadata presented. See the paper I link to in a previous comment.

daniel-beck:
> Currently limited to 2 versions per plugin, probably because of an earlier performance improvement I cannot find right now. Will need to dig a bit.

Implemented in `VersionCappedMavenRepository` for some reason, will not be a problem on the live site since we only do plugin versions on the uncapped run; adapted Jenkinsfile to also have such a run (which may be a terrible idea, PR verification wise).

daniel-beck:
@ndeloof I finally remembered why there's a separate `plugins` element: To allow the entire file to contain a signature, similar to `update-center.actual.json`. It's not available in the PR build as that is unsigned, but it would be a sibling element to `plugins` and `updateCenterVersion` (which really should just be removed).

daniel-beck:
This approach is impractical as it undoes the performance improvements from many earlier PRs that do not require us to download every release. PR builds aren't getting finished in hours, or even days.

ndeloof:
@daniel-beck IIUC the issue is that PR CI would require to build a full update center and doing so would run for ages ? Can't you add an option to limit to `x` release per plugin so we check this feature isn't impacted ?

ndeloof:
I misunderstood your comment, thought the issue was for building PRs, not for the update center cron job.
I wonder : doesn't this job download ALL plugins and versions already ? As they are hosted, I was assuming they are downloaded during the UC build process. Or maybe is this incremental only ?

What about having metadata file _per_ plugin+version (as I initially suggested) aside artifact, and this plugins-versions.json file only provide a link ? Doing so you can generate this later file without actually downloading plugins, and just need to run a full Update Center build once to get per-pluing-and-version metadata generated once.

daniel-beck:
> I misunderstood your comment, thought the issue was for building PRs, not for the update center cron job.

I specifically patched the Jenkinsfile to have a more realistic full run once I noticed PR builds are only doing the optimized run.

>  I wonder : doesn't this job download ALL plugins and versions already ? As they are hosted, I was assuming they are downloaded during the UC build process. Or maybe is this incremental only ?

It creates hard links to the local Maven repo in the output, so that it can then be rsynched out.

Good point. So perhaps this is a problem for PR builds only, and the extreme duration was a red herring? I will need to investigate this some more, perhaps with more verbose logging to see exactly what is happening.

> and this plugins-versions.json file only provide a link ? Doing so you can generate this later file without actually downloading plugins, and just need to run a full Update Center build once to get per-pluing-and-version metadata generated once.

If there's no secure link (referenced file's checksum) in the global file a number of vulnerabilities are enabled, so we cannot just link into a void, to be filled later. Once again, please read the paper on update repository security I linked to before.

If the above with the actual infra having no extreme duration problem does not hold true, we may be able to do something like asynchronously backfill metadata files for individual downloads as needed in a separate process, and, if they exist on update center generation, just link them from the main update sites, which *should* be lightweight enough to accomplish your goals within "my" constraints -- although with some delay.

ndeloof:
created https://issues.jenkins-ci.org/browse/INFRA-1696 to track this

daniel-beck:
I'm merging this as it appears to be side effect free, and am going to monitor the build durations. Given what the production update center build currently does, there's a good chance this won't impact those too much.

daniel-beck:
Well, this is bad, we're up to >1 hour from 20 minutes before.

Looking into updating this entire thing to maven-indexer 6.0.0 to get sha256 in the artifact metadata, so it doesn't have to be computed.

ndeloof:
Can't you use Artifactory API to request sha256 and other such things based on GAV ?

daniel-beck:
@ndeloof That would be the next option, but going the naive approach is still tens of thousands of HTTP requests, so I don't know how much quicker that would be. So it'd involve requesting and processing a 30 MB JSON file list. That's probably the best alternative.

daniel-beck:
Sadly I went full SHA-512 when I added better checksums recently, so that may need a corresponding change to core to scale it back a bit.

PR #204, opened 2018-06-15
 [INFRA-1666] Prefer index.jelly to <description> for excerpt

jglick:
[INFRA-1666](https://issues.jenkins-ci.org/browse/INFRA-1666)

See [installed view](https://github.com/jenkinsci/jenkins/blob/6c49fabc45062e86f6243225a4c6c6a929725fc0/core/src/main/resources/hudson/PluginManager/installed.jelly#L87-L100) vs. [update view](https://github.com/jenkinsci/jenkins/blob/6c49fabc45062e86f6243225a4c6c6a929725fc0/core/src/main/resources/hudson/PluginManager/table.jelly#L100-L102).

Before, from [this](https://github.com/jenkinsci/accurev-plugin/blob/b2a01a82c982f61bc50e96af548e339f9c29ff61/pom.xml#L15):

```json
"excerpt": "This plugin allows you to use AccuRev as a SCM.",
```

After, from [this](https://github.com/jenkinsci/accurev-plugin/blob/b2a01a82c982f61bc50e96af548e339f9c29ff61/src/main/resources/index.jelly#L3):

```json
"excerpt": "This plugin integrates <a href=\"http://www.accurev.com/\">AccuRev SCM<\/a> to Jenkins."
```

More to the point, before incrementalification:

```json
"excerpt": ""
```

and after incrementalification:

```json
"excerpt": "The Jenkins Plugins Parent POM Project"
```

and after this PR:

```json
"excerpt": "Plugin that defines Pipeline API."
```

CC @daniel-beck @kshultzCB

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,12 @@
       <groupId>org.codehaus.plexus</groupId>
       <artifactId>plexus-container-default</artifactId>
       <version>1.5.5</version>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.collections</groupId>
+          <artifactId>google-collections</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.wagon</groupId>
@@ -173,6 +179,11 @@
       <groupId>jetty</groupId>
       <artifactId>jetty-util</artifactId>
       <version>6.0.0rc1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
+      <artifactId>owasp-java-html-sanitizer</artifactId>
+      <version>20180219.1</version>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
--- a/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
+++ b/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
@@ -110,6 +110,14 @@ public class MavenArtifact {
         }
     }
 
+    public File resolveJar() throws IOException {
+        try {
+            return repository.resolve(artifact, "jar", null);
+        } catch (AbstractArtifactResolutionException e) {
+            throw new IOException("Failed to resolve artifact " + artifact + " jar", e);
+        }
+    }
+
     public File resolveSources() throws IOException {
         try {
             return repository.resolve(artifact,"jar","sources");
--- a/src/main/java/org/jvnet/hudson/update_center/MavenRepositoryImpl.java
+++ b/src/main/java/org/jvnet/hudson/update_center/MavenRepositoryImpl.java
@@ -66,6 +66,7 @@ import java.io.IOException;
 import java.net.URL;
 import java.net.URLConnection;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
@@ -161,7 +162,7 @@ public class MavenRepositoryImpl extends MavenRepository {
 
         URLConnection con = url.openConnection();
         if (url.getUserInfo()!=null) {
-            con.setRequestProperty("Authorization","Basic "+new sun.misc.BASE64Encoder().encode(url.getUserInfo().getBytes("UTF-8")));
+            con.setRequestProperty("Authorization", "Basic " + Base64.getEncoder().encodeToString(url.getUserInfo().getBytes("UTF-8")));
         }
 
         if (!expanded.exists() || !local.exists() || (local.lastModified()!=con.getLastModified() && !offlineIndex)) {
--- a/src/main/java/org/jvnet/hudson/update_center/Plugin.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Plugin.java
@@ -52,6 +52,8 @@ import java.util.logging.Logger;
 import java.util.regex.Pattern;
 import java.util.zip.ZipEntry;
 import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamResult;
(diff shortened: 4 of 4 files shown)


daniel-beck:
Also, before:

    "excerpt": "",

After:

    "excerpt": "<f:entry title=\"TestSwarm Server URL\">\n\t\t<f:textbox checkMessage=\"${%mandatory.testswarmServerUrl}\" clazz=\"required\" field=\"testswarmServerUrl\" value=\"${instance.testswarmServerUrl}\"/>\n  \t</f:entry>\n  \t  \t\n  \t<f:entry title=\"Job Name\">\n\t\t<f:textbox checkMessage=\"${%mandatory.jobName}\" clazz=\"required\" default=\"Hudson - Integration Tests\" field=\"jobName\" value=\"${instance.jobName}\"/>\n  \t</f:entry>\n  \t\n  \t<f:entry title=\"Your Username\">\n\t\t<f:textbox checkMessage=\"${%mandatory.userName}\" clazz=\"required\" default=\"testswarm\" field=\"userName\" value=\"${instance.userName}\"/>\n  \t</f:entry>\n  \t\n  \t<f:entry title=\"Your Auth Token\">\n\t\t<f:textbox checkMessage=\"${%mandatory.authToken}\" clazz=\"required\" default=\"password\" field=\"authToken\" value=\"${instance.authToken}\"/>\n  \t</f:entry>   \n  \t\n  \t<f:entry title=\"Number of Runs\">\n\t\t<f:textbox checkMessage=\"${%mandatory.maxRuns}\" clazz=\"required\" default=\"3\" field=\"maxRuns\" value=\"${instance.maxRuns}\"/>\n  \t</f:entry>\n  \t\n\t<f:entry field=\"chooseBrowsers\" name=\"chooseBrowsers\" title=\"Choose Browsers\">\n    \t<select name=\"chooseBrowsers\">\n            <option value=\"popular\">Supported browsers</option>\t    \t\n        \t<option value=\"all\">All available browsers</option>\n            <option value=\"current\">Current release of all major browsers</option>\n            <option value=\"gbs\">Browsers currently supported in Yahoo's Graded Browser Support</option>\n            <option value=\"beta\">Upcoming alphas/betas of popular browsers</option>\n            <option value=\"popularbeta\">Popular browsers and their upcoming alphas/betas</option>\n        </select>\n    </f:entry>   \n\n  \t<f:entry title=\"Polling Interval (in secs)\">\n\t\t<f:textbox checkMessage=\"${%mandatory.pollingIntervalInSecs}\" clazz=\"required\" default=\"10\" field=\"pollingIntervalInSecs\" value=\"${instance.pollingIntervalInSecs}\"/>\n  \t</f:entry>\n  \t\n  \t<f:entry title=\"Timeout Period (in mins)\">\n\t\t<f:textbox checkMessage=\"${%mandatory.timeOutPeriodInMins}\" clazz=\"required\" default=\"2\" field=\"timeOutPeriodInMins\" value=\"${instance.timeOutPeriodInMins}\"/>\n  \t</f:entry>  \t  \t\n\t\n\t<f:entry title=\"Test Suite\">\n      <f:repeatable items=\"${instance.testSuiteList}\" minimum=\"1\" name=\"testSuiteList\" noAddButton=\"true\" var=\"testSuiteData\">\n        <table width=\"100%\">\n          <br/>\n          <f:entry field=\"testName\" title=\"Name\">\n            <f:textbox value=\"${testSuiteData.testName}\"/>\n          </f:entry>\n          <f:entry field=\"testUrl\" title=\"URL\">\n            <f:textbox value=\"${testSuiteData.testUrl}\"/>\n          </f:entry>          \n\t\t  <f:entry field=\"testCacheCracker\" title=\"Enable CacheCracker\">\n            <f:checkbox checked=\"${testSuiteData.testCacheCracker}\" default=\"true\" name=\"testCacheCracker\"/>\n          </f:entry>\n          <f:entry field=\"disableTest\" title=\"Disable Test\">\n            <f:checkbox checked=\"${testSuiteData.disableTest}\" name=\"disableTest\"/>\n          </f:entry>\n\t\t<f:entry>\n        <div align=\"right\">\n          <input class=\"repeatable-add show-if-last\" type=\"button\" value=\"Add more test suites...\"/>\t\n          <input class=\"repeatable-delete show-if-not-only\" style=\"margin-left: 1em;\" type=\"button\" value=\"Delete\"/>\n        </div>\n        </f:entry>\n        </table>\n      </f:repeatable>          \n\t</f:entry>",

Not sure this is always an improvement 😆

jglick:
> Needs HTML sanitization.

Can you expand on what you are expecting?

> Jenkins […] trusts the update center to not provide malicious descriptions.

I suppose you mean that even though we trust plugin authors to not supply malicious _code_, we do not trust them to not supply malicious _descriptions_ to people who have not installed the plugin? OK in that case we would need to run the OWASP sanitizer or something. I can take a stab at it.

jglick:
BTW for the example you mentioned: https://github.com/jenkinsci/testswarm-plugin/pull/1 was merged over five years ago but never released. Perhaps this plugin should be dropped from the update center.

daniel-beck:
> Can you expand on what you are expecting?

Right now, Jenkins admins have to trust the developers of plugins they are installing. This is, within certain bounds, doable. Worst case, just code-review everything before you install it. This is pretty much inherent in community provided update sites.

With this change, Jenkins admins have to trust all developers of any published Jenkins plugin not to provide malicious descriptions.

daniel-beck:
Grows output by ~60KB or about 4% in my review, which should be acceptable. Pinging @rtyler anyway.

daniel-beck:
Notably, there are currently 400+ plugins that get a description that used to be empty. Only ~30 seem to have unreasonable descriptions in the file (unchanged from the sample plugin, or something like `${%PluginDescription}`), so it seems like a good net positive based on that alone (and there seems to be quite some overlap between these).

jglick:
If accepted, we should also be able to delete the `<description>` from the plugin archetype, so plugin authors are not bothered by the need to maintain two copies of the same text.

batmat:
I am fine with this change, but somehow this gets me thinking we should JEP this to make it durable: documented and agreed on.

jglick:
A JEP feels like overkill for this. As far as I am concerned it is just a bug fix: the **Available**/**Updates** tabs sometimes displayed different content from the **Installed** tab (when the content did not actually change from version to version, that is!), due to improper behavior in the UC generator.

batmat:
@daniel-beck I guess we can merge this? Thanks

batmat:
About the JEP, I meant more a general one that would include this part. Not specifically for this only.

daniel-beck:
Easy to undo in the case this is an undue burden on infra.

jglick:
`/pluginManager/available` on a fresh installation looks good to me now.

daniel-beck:
Temporarily reverted this to restore the jenkins.io site build. The update center metadata is read as YAML and processed during the site build to make the metadata available on the site (for e.g. the names of plugins listed in security advisories), but the richer descriptions mess this up:

    An error occurred: (…/jenkins.io.git/content/_data/_generated/update_center.yml): found unknown escape character while parsing a quoted scalar at line 1 column 3152

Specifically it seems to object to `\"` being used to escape quotes around HTML attribute values.

Not sure how to address this without writing more code than I'd like for the site.

jglick:
Is there a simple known way to reproduce this error locally with a `jenkins.io` checkout? I can try to find a fix for it.

daniel-beck:
@jglick Dump the output of this into `content/_data/_generated/update-center.yml`.

jglick:
Thanks. Will try to work on this shortly.

PR #205, opened 2018-06-18
Hide a release with no docs and released outside from jenkinsci org

batmat:


daniel-beck:
This is from before upload permissions as per https://github.com/jenkins-infra/repository-permissions-updater/blob/master/permissions/plugin-scalable-amazon-ecs.yml

It's not clear to me what distinguishes this from other plugins we have grandfathered in. The fork relationship to amazon-ecs?

daniel-beck:
Link to tag: https://github.com/roehrijn/amazon-ecs-plugin-roehrijn-fork/tree/1.0

batmat:
Forgot to add the reason I think it's not normal: it was 6+ months already after we introduced the dedicated jira HOSTING project.

Even this, not the first one, https://issues.jenkins-ci.org/browse/HOSTING-12 predates that (imo) rogue release by 6 months.

PR #206, opened 2018-07-11
Remove SHA-512 support, use Artifactory API to obtain SHA-1/256

daniel-beck:
Followup to #198. Companion to jenkinsci/jenkins#3546. Mostly [INFRA-1696](https://issues.jenkins-ci.org/browse/INFRA-1696), with [INFRA-1503](https://issues.jenkins-ci.org/browse/INFRA-1503) and [INFRA-1666](https://issues.jenkins-ci.org/browse/INFRA-1666) on the side.

Adds a requirement for the environment to have `ARTIFACTORY_USERNAME`/`ARTIFACTORY_PASSWORD` to run, otherwise falls back to current approach of checksumming local files.

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/IndexHtmlBuilder.java
+++ b/src/main/java/org/jvnet/hudson/update_center/IndexHtmlBuilder.java
@@ -90,9 +90,7 @@ public class IndexHtmlBuilder implements Closeable {
     }
 
     public void add(String url, String caption) throws MalformedURLException {
-        out.println(
-                "<tr><td><img src='http://jenkins-ci.org/images/jar.png'/></td><td><a href='"+ url +"'>"+ caption +"</a></td></tr>"
-        );
+        add(url, caption, null);
     }
 
     public void add(String url, String caption, String metadata) throws MalformedURLException {
--- a/src/main/java/org/jvnet/hudson/update_center/IndexHtmlBuilder.java
+++ b/src/main/java/org/jvnet/hudson/update_center/IndexHtmlBuilder.java
@@ -23,6 +23,9 @@
  */
 package org.jvnet.hudson.update_center;
 
+import org.apache.commons.codec.binary.Hex;
+import org.bouncycastle.util.encoders.Base64;
+
 import java.io.Closeable;
 import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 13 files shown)


daniel-beck:
Screenshot for https://github.com/jenkins-infra/update-center2/pull/206/commits/358abd1792bfbf7cb1ff9e5923343679d429f682:

screen shot

Right now, only artifacts uploaded since October 1, 2017 have SHA-256. See comments in INFRA-1696.

daniel-beck:
http://lists.jenkins-ci.org/pipermail/jenkins-infra/2018-July/001472.html 🎉

casz:
This deserves a

src/main/java/org/jvnet/hudson/update_center/MavenArtifactSource.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    @Override
+    public Manifest getManifest(MavenArtifact artifact) throws IOException {
+        Manifest manifest;
+        File f = artifact.resolve();
+        try {
+            JarFile jar = new JarFile(f);
+            ZipEntry e = jar.getEntry("META-INF/MANIFEST.MF");
```
jglick: unused

daniel-beck: Inherited from https://github.com/jenkins-infra/update-center2/blob/65636cef1cef2919057e78f09664e3b9b3820a01/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java#L203 🤷‍♂️

src/main/java/org/jvnet/hudson/update_center/IndexHtmlBuilder.java
```diff
@@ -69,16 +72,41 @@ public IndexHtmlBuilder(PrintWriter out, String title) {
...
     public void add(String url, String caption) throws MalformedURLException {
         out.println(
-            "<tr><td><img src='http://jenkins-ci.org/images/jar.png'/></td><td><a href='"+ url +"'>"+ caption +"</a></td></tr>"
+                "<tr><td><img src='http://jenkins-ci.org/images/jar.png'/></td><td><a href='"+ url +"'>"+ caption +"</a></td></tr>"
         );
     }
 
+    public void add(String url, String caption, String metadata) throws MalformedURLException {
+        if (metadata == null) {
+            out.println(
```
casz: Could just call `add(url, caption)` instead of duplicating 😄

daniel-beck: I think I meant to make `add(String, String)` call this with a `null` third arg. Unsure. This went through a few iterations as I wasn't sure whether the third arg should bring its `<td>` wrapper or not.

daniel-beck:
@jglick Any reason not to just do this like https://gist.github.com/daniel-beck/a776ea32b537cf0b12cf699cd12b0647 ?

Checking output differences. 21 descriptions are affected, all but one of them due to containing `<div>`, `<p>`, `<ul>`/`<li>`. The last one has some broken HTML that gets rendered as written: https://github.com/jenkinsci/performance-plugin/blob/40ab8fecbe84839b17311365e1ad13571f5f7153/src/main/resources/index.jelly#L11

I would actually consider the lack of blocks like this to be preferable.

jglick:
> Any reason not to just do this like …?

If it works, sure. I guess you are basically overloading the HTML sanitizer to strip out `j:jelly` and the like.

daniel-beck:
> basically overloading the HTML sanitizer to strip out

Yes, exactly. The output is identical except for what I identified above when I tried this. Will propose a PR. Thanks @jglick

PR #217, opened 2018-07-19
[INFRA-1666] Prefer index.jelly to <description> for excerpt (redux)

jglick:
Redoing #204, with a few additional refinements.

I have been trying to reproduce the claimed `jenkins.io` error with `\"` without success. I generated `actual.json` with `-maxPlugins 100`, enough to include some plugins with the supposedly problematic sequence:

```json
"excerpt": "This plugin integrates <a href=\"http://www.accurev.com/\" rel=\"nofollow\">AccuRev SCM</a> to Jenkins."
```

and then copied it to `content/_data/_generated/update_center.yml`. I tried

```bash
make
```

which worked, as did

```bash
make run
```

when I browsed, for example, `http://localhost:4242/security/advisory/2018-04-16/`, based on grepping sources and finding apparent usages in `content/security/advisories.html.haml` & `content/_layouts/advisory.html.haml`.

I suppose the expectation is that, as in [this hint](https://stackoverflow.com/a/24710810/12916), well-formed JSON can be parsed as YAML. Indeed in `irb`

```ruby
require 'yaml'
puts YAML.load('{"excerpt": "Invoke <a href=\"http://www.ansible.com/\" rel=\"nofollow\">Ansible</a> Ad-Hoc commands and playbooks."}')['excerpt']
```

works as expected.

How precisely do I reproduce the stated error?

Presumably the fix will somehow involve making `Makefile` and/or `scripts/fetch-external-resources` a little smarter: rather than copying `update-center.actual.json` verbatim to `update_center.yml`, we could process it a bit with `jq` or whatever to retrieve only the fields we are actually using (just `plugins/*/title` apparently), and ideally convert to regular YAML format.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -153,6 +153,12 @@
       <groupId>org.codehaus.plexus</groupId>
       <artifactId>plexus-container-default</artifactId>
       <version>1.5.5</version>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.collections</groupId>
+          <artifactId>google-collections</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.wagon</groupId>
@@ -183,6 +189,11 @@
       <groupId>jetty</groupId>
       <artifactId>jetty-util</artifactId>
       <version>6.0.0rc1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
+      <artifactId>owasp-java-html-sanitizer</artifactId>
+      <version>20180219.1</version>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
--- a/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
+++ b/src/main/java/org/jvnet/hudson/update_center/MavenArtifact.java
@@ -110,6 +110,14 @@ public class MavenArtifact {
         }
     }
 
+    public File resolveJar() throws IOException {
+        try {
+            return repository.resolve(artifact, "jar", null);
+        } catch (AbstractArtifactResolutionException e) {
+            throw new IOException("Failed to resolve artifact " + artifact + " jar", e);
+        }
+    }
+
     public File resolveSources() throws IOException {
         try {
             return repository.resolve(artifact,"jar","sources");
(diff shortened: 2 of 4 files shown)


daniel-beck:
Please rebase to current master to fix archiving of generated files.

jglick:
OK, can finally reproduce, thanks.

jglick:
Merge https://github.com/jenkins-infra/jenkins.io/pull/1665 first and we should be good.

daniel-beck:
This will need to be reworked to use the Artifactory API and cache implemented in https://github.com/jenkins-infra/update-center2/pull/206 after merge. Accessing all the artifacts locally will just be too expensive with the full history.

Not sure I want to merge this independently of the other PR (which is tentatively scheduled for next week once JFrog upgraded the index to include SHA-256).

jglick:
…though it will be a bit messy until the refactoring from #206 allows us to introduce a new method in `ArtifactSource` to load `index.jelly`. Or better, a more abstract set of methods to load JAR entries and sub-JAR entries.

daniel-beck:
> Could be reworked that way now if you like.

Yes, but with the other PR being tied to infra configuration changes for credentials, as well as the SHA-256 change that will only be available from next week, the (minor) rework is blocked by the other PR.

Might just merge this into the other branch and rewrite these parts to use `ArtifactoryArtifactSource`.

jglick:
> the other PR being tied to infra configuration changes for credentials

Again note FWIW that what _this_ PR needs from Artifactory (to avoid downloading large binaries) does not seem to require any credentials. At any rate, it _would_ best be done atop the API refactoring done there.

daniel-beck:
> Again note FWIW that what this PR needs from Artifactory (to avoid downloading large binaries) does not seem to require any credentials. At any rate, it would best be done atop the API refactoring done there.

Unless this is much more urgent than I expect (if so please let me know) I doubt the extra few days of wait time to integrate this with the other change for release next week shouldn't be worth pulling the API additions apart.

daniel-beck:
Subsumed into #206 via https://github.com/jenkins-infra/update-center2/pull/206/commits/2dcd9076918c93d2f763106835db45a4b8edd650, amended by https://github.com/jenkins-infra/update-center2/pull/206/commits/a2561bce11c624af88336f6abd3a2d80eec5dc03

daniel-beck:
FWIW I expect SHA-256 reindexing to be done round Monday/Tuesday midnight UTC. So mid this week, we can probably go ahead with #206.

PR #236, opened 2018-09-04
[JENKINS-53399] Blacklist config-file-provider 3.0

abayer:
This breaks too badly - see
https://issues.jenkins-ci.org/browse/JENKINS-53399. I just helped
someone deal with the result of upgrading from 2.18 to 3.0 - folder
shared libraries and permissions all broke horribly, and were fixed
with a downgrade.


oleg-nenashev:
Note that 3.0 is 3.0, it would just need compatibleSince and documentation.
CC @imod anyway

abayer:
Build passed, Oleg said yes, I have merge rights, so...merging.

PR #238, opened 2018-09-20
Suspend distribution of sge-cloud-plugin

daniel-beck:
The first actual release of `sge-cloud-plugin` was performed this week.

Unfortunately, it has a mandatory dependency on [the copy-to-slave plugin we blacklisted in March](https://jenkins.io/security/advisory/2018-03-26/#SECURITY-545), so it cannot actually be installed. Additionally, Jenkins instances will show this warning, even if they have neither mentioned plugin installed:

    WARNING: Could not find dependency copy-to-slave of sge-cloud-plugin

Suspending distribution until this dependency is removed or made optional.

FYI @jmcgeheeiv

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -269,6 +269,7 @@ liquibase-runner
 perforce
 build-configurator  # depends on copy-to-slave
 lsf-cloud           # depends on copy-to-slave
+sge-cloud-plugin    # depends on copy-to-slave
 reviewboard         # depends on perforce
 
 # requested by maintainer in https://github.com/jenkins-infra/update-center2/pull/199


jmcgeheeiv:
This looks real to me.  I'll get to work on it.

Since I have your attention, I apologize for interrupting your conversation last night.  I just wanted to thank @daniel-beck before I lost my chance.

jmcgeheeiv:
Regarding the solution, would it be satisfactory to incorporate the relevant portion of `copy-to-slave` into `sge-cloud-plugin` such that `sge-cloud-plugin` controls the exact files copied?

daniel-beck:
@jmcgeheeiv 

> would it be satisfactory

Should be OK, but devil's in the details in something like this. Also possibly a licensing issue, try to get that right. Beyond that, happy to review the impl, ideally submitted as a PR to your plugin.

PR #243, opened 2018-10-05
[JENKINS-53917] Remove structs 1.16

daniel-beck:
[JENKINS-53917](https://issues.jenkins.io/browse/JENKINS-53917)

Not sure who would make the decision to pull this, but prepared it just in case.


daniel-beck:
Superseded.

PR #245, opened 2018-10-15
[INFRA-1809] Private update center improvement

LinuxSuRen:
[INFRA-1809](https://issues.jenkins-ci.org/browse/INFRA-1809)

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,3 +110,31 @@ you can try to use the appassembler plugin as described below. The exec:java plu
     # warning this may take quite a bit of time, so you might want to add the -maxPlugins 1 option
     mvn package appassembler:assemble
     sh target/appassembler/bin/app -id com.example.jenkins -www www
+
+Private Update Center
+---------------------
+
+Jenkins offical update center has lots of plugins. You could follow below steps to create your own
+private update center:
+
+1. Create your own certificate. For example:
+
+    openssl genrsa -out rootCA/demo.key 1024
+    openssl req -new -x509 -days 1095 -key rootCA/demo.key \
+        -out rootCA/demo.crt \
+        -subj "/C=CN/ST=GD/L=SZ/O=vihoo/OU=dev/CN=demo.com/emailAddress=demo@demo.com"
+2. Fetch plugins information then generate update.json
+
+    echo "localization-zh-cn=1.580.1" > whiteList.properties
+    mvn package appassembler:assemble
+	sh target/appassembler/bin/app -id default -www www \
+		-skip-release-history -cache plugins -whitelist whiteList.properties \
+		-key rootCA/demo.key -certificate rootCA/emo.crt \
+		-root-certificate rootCA/demo.crt \
+		-cacheServer http://localhost:9090/plugins/ \
+		-connectionCheckUrl http://localhost:9090/
+3. Start your update center server (e.g. Nginx). Copy `www` into publish directory.
+4. Start your Jenkins server then copy demo.crt into `$JENKINS_HOME/war/WEB-INF/update-center-rootCAs/`.
+5. Change the update center url in `http://localhost:8080/pluginManager/advanced`.
+
+Finally, you could see your favour plugins.
--- a/README.md
+++ b/README.md
@@ -119,20 +119,24 @@ private update center:
 
 1. Create your own certificate. For example:
 
-    openssl genrsa -out rootCA/demo.key 1024
-    openssl req -new -x509 -days 1095 -key rootCA/demo.key \
-        -out rootCA/demo.crt \
-        -subj "/C=CN/ST=GD/L=SZ/O=vihoo/OU=dev/CN=demo.com/emailAddress=demo@demo.com"
+```shell script
+openssl genrsa -out rootCA/demo.key 1024
+openssl req -new -x509 -days 1095 -key rootCA/demo.key \
+    -out rootCA/demo.crt \
+    -subj "/C=CN/ST=GD/L=SZ/O=vihoo/OU=dev/CN=demo.com/emailAddress=demo@demo.com"
+```
 2. Fetch plugins information then generate update.json
 
-    echo "localization-zh-cn=1.580.1" > whiteList.properties
-    mvn package appassembler:assemble
-	sh target/appassembler/bin/app -id default -www www \
-		-skip-release-history -cache plugins -whitelist whiteList.properties \
-		-key rootCA/demo.key -certificate rootCA/demo.crt \
-		-root-certificate rootCA/demo.crt \
-		-cacheServer http://localhost:9090/plugins/ \
-		-connectionCheckUrl http://localhost:9090/
+```shell script
+echo "localization-zh-cn=" > whiteList.properties
+mvn package appassembler:assemble
+sh target/appassembler/bin/app -id default -www www \
+    -skip-release-history -cache plugins -whitelist whiteList.properties \
+    -key rootCA/demo.key -certificate rootCA/emo.crt \
+    -root-certificate rootCA/demo.crt \
+    -cache-server http://localhost:9090/plugins/ \
+    -connectionCheckUrl http://localhost:9090/
+```
 3. Start your update center server (e.g. Nginx). Copy `www` into publish directory.
 4. Start your Jenkins server then copy demo.crt into `$JENKINS_HOME/war/WEB-INF/update-center-rootCAs/`.
 5. Change the update center url in `http://localhost:8080/pluginManager/advanced`.
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ private update center:
     mvn package appassembler:assemble
 	sh target/appassembler/bin/app -id default -www www \
 		-skip-release-history -cache plugins -whitelist whiteList.properties \
-		-key rootCA/demo.key -certificate rootCA/emo.crt \
+		-key rootCA/demo.key -certificate rootCA/demo.crt \
 		-root-certificate rootCA/demo.crt \
 		-cacheServer http://localhost:9090/plugins/ \
 		-connectionCheckUrl http://localhost:9090/
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -96,7 +96,7 @@ public class Main {
     @Option(name="-cache",usage="Cache directory used for caching plugin files")
     public File cache = null;
 
-    @Option(name="-cacheAll",usage="Cache all version of plugin")
+    @Option(name="-cacheAll",usage="Cache all version of plugin instead of only the latest")
     public boolean cacheAll;
 
     @Option(name="-cacheServer",usage="Cache server will replace the orignal")
--- a/src/main/java/org/jvnet/hudson/update_center/Main.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Main.java
@@ -96,10 +96,10 @@ public class Main {
     @Option(name="-cache",usage="Cache directory used for caching plugin files")
     public File cache = null;
 
-    @Option(name="-cacheAll",usage="Cache all version of plugin instead of only the latest")
+    @Option(name="-cache-all",usage="Cache all version of plugin instead of only the latest")
     public boolean cacheAll;
 
-    @Option(name="-cacheServer",usage="Cache server will replace the orignal")
+    @Option(name="-cache-server",usage="Cache server will replace the orignal")
     public String cacheServer;
 
     /**
@@ -269,7 +269,7 @@ public class Main {
         // load white list of plugins
         if(whitelist != null && whitelist.isFile()) {
             try(InputStream input = new FileInputStream(whitelist)) {
-                whitelistPro.load(input);
+                whitelistProperties.load(input);
             }
         }
 
@@ -389,7 +389,7 @@ public class Main {
         System.out.println("Build plugin versions index from the maven repo...");
 
         for (PluginHistory plugin : repository.listHudsonPlugins()) {
-                if(whitelistPro.size() > 0 && whitelistPro.get(plugin.artifactId) == null) {
+                if(whitelistProperties.size() > 0 && whitelistProperties.get(plugin.artifactId) == null) {
                     continue;
                 }
                 System.out.println(plugin.artifactId);
@@ -447,7 +447,7 @@ public class Main {
         }
         System.out.println("Gathering list of plugins and versions from the maven repo...");
         for (PluginHistory hpi : repository.listHudsonPlugins()) {
-            if(whitelistPro.size() > 0 && whitelistPro.get(hpi.artifactId) == null) {
+            if(whitelistProperties.size() > 0 && whitelistProperties.get(hpi.artifactId) == null) {
                 continue;
             }
 
@@ -616,7 +616,7 @@ public class Main {
                 JSONObject o = new JSONObject();
                 try {
                     Plugin plugin = new Plugin(h);
-                    if(whitelistPro.size() > 0 && whitelistPro.get(plugin.artifactId) == null) {
+                    if(whitelistProperties.size() > 0 && whitelistProperties.get(plugin.artifactId) == null) {
                         continue;
                     }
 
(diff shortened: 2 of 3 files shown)


daniel-beck:
Features not exercised in Jenkins project infra are unlikely to be maintained well.

Note that this _will not_ be merged, as it has content specific to your environment, rather than just generalizations (which could be). It seems you should rather maintain a fork with your specific data, rather than seek merging into upstream.

Wouldn't https://github.com/yandex-qatools/juseppe be better for your use case?

LinuxSuRen:
I think there is a misunderstanding. I just find I commit some private files. That's no need. I will remove them.
So I mean this PR will don't contain any specific environment information.

LinuxSuRen:
Okay, just miss this.

rasmusjelsgaard:
We have the same use case where I work as @LinuxSuRen has described as the motivation for this PR.
It seems it has been waiting for some time. Anything I could do to help to speed up things? Or is there anything preventing this from being merged?

LinuxSuRen:
ping @daniel-beck How do you think about this PR? Thanks for your time.

LinuxSuRen:
Anyone could help me to review this? Thanks.

daniel-beck:
I recommend you maintain a fork of this repository if you want to use it on your own infrastructure.

It's not a goal of this tool to be generally useful, and if we have to cut corners, or even overhaul the tool, to optimize it for use in the Jenkins project infrastructure, we will.

I'll poke Tyler to see whether he disagrees, but I'm not particularly inclined to add features (and the resulting risk of getting bug reports and RFEs I have no intention in working on) that would not be exercised in Jenkins infra.

LinuxSuRen:
@daniel-beck I can understand what you're thinking. And I agree with you that this org (jenkins-infra) is setting up for Jenkins infrastructure. I can maintain this feature. But I believe that there're lots of people have the same demands as me. Anyway, thanks for your viewpoint.

daniel-beck:
@LinuxSuRen Thanks for your understanding. I've talked with others in the past and they mentioned their problems in keeping their installations up to date with this repo, and their specific needs (like ones implemented here), so I understand the use case. It's just something we don't really have the capacity to support. As I wrote I will check in with @rtyler before closing this PR in case he offers a different perspective.

A real fork (not like GitHub forks, more like Hudson/Jenkins) of this repo for end user update sites, or perhaps an entirely different tool like https://github.com/jenkinsci/juseppe, seems like a much easier way to handle this use.

rtyler:
@daniel-beck I definitely understand your concerns and share many of them myself. If I look at [pipeline-library](https://github.com/jenkins-infra/pipeline-library) however, there has been a fair bit of work done to make that library work _outside_ of the Jenkins project's own CI infrastructure. 

In that case, the iteration on the code for uses not strictly part of the original intent for pipeline-library have proven to be overall beneficial to the usage on ci.jenkins.io.

I think if @LinuxSuRen is volunteering to contribute and improve the code in this repository, that is an overall long term benefit to the update center process running for the Jenkins project, even if it may seem unrelated in the short term. 

Fortunately, we know @LinuxSuRen so I'm not worried about this being a "drive-by contribution." :smile_cat:

Siegfriedk:
We have the same issue: For security reasons and building jenkins images on the fly, we are very dependend on it and need an internal mirror.

The code itself is not much and it looks very maintainable. What would make it possible to integrate it in this repo? Any official commitment?

I have found a few ways of how people are simulating what this tool is doing. Not sure whats more beneficial for jenkins-infra: Some hacky way which is not controlable or an official one :).

I will clone it and work with LinuxSuRen branch. @LinuxSuRen tx for it

LinuxSuRen:
I really hope this official one could also support a private environment. At least, this feature would not break any infra I think.

And Thanks @Siegfriedk and @rasmusjelsgaard to using this PR.

reenberg:
Awesome, I was actually in the process of implementing something like this.

However I was also thinking of implementing a limit on how many plugins and war files to keep around.  It seems a bit silly to offer every version of every plugin inside a closed network.

Siegfriedk:
> Awesome, I was actually in the process of implementing something like this.
> 
> However I was also thinking of implementing a limit on how many plugins and war files to keep around. It seems a bit silly to offer every version of every plugin inside a closed network.

I thought about this as well but hadn't time to look into it. I would probably add a filter like "last 5 versions" or "younger then 5 month"

For now, with the cache, it works quite well and when you look into the setup / content of all plugins there are not tooo many which are huge.

For example, a quick check on one of our older jenkins mirror setups:
`483M	selenium
497M	docker-build-step
524M	azure-vm-agents
572M	dependency-check-jenkins-plugin
579M	ontrack
597M	doktor
609M	openshift-sync
623M	cucumber-living-documentation
628M	findbugs
659M	pmd
681M	yet-another-docker-plugin
798M	kubernetes
810M	xframium
882M	aws-codepipeline
888M	DotCi
950M	crossbrowsertesting
1.2G	artifactory
1.2G	aws-device-farm
1.2G	aws-java-sdk
1.3G	whitesource
2.0G	maven-plugin
7.5G	sauce-ondemand`

reenberg:
> I thought about this as well but hadn't time to look into it. I would probably add a filter like "last 5 versions" or "younger then 5 month"
Just my idea.

> For now, with the cache, it works quite well and when you look into the setup / content of all plugins there are not tooo many which are huge.

Size was one thing, but also sync time was a concern for me.  Lastly i was thinking about reducing the load on the jenkins-ci infrastructure.  
Note that I haven't completely looked into how much network traffic it generates when you already have synched it once, but I have noticed that it takes quite some time.

reenberg:
For what it is worth, I have tried giving it a stab.

I have forked master and rebased the changes made by @LinuxSuRen.  However they had some issues.  For example, it was only the plugins that had their url changed in the produced `update-center.json` file, the .war artifacts was still using the original hard coded url.

Since the whole concept of a cache server didn't make sense to me, as the `-download` option will produce the needed artifacts for you, and since you can now change the URL, I decided to more or less remove all the additions that was made, and then I renamed the parameter for setting the url, to something that I found more natural.

In order to clean up some code, I decided to implement a static `Options` class, that contains all the cli parameters.  This made it possible to read those parameters from the `HPI` and `HudsonWar` classes, which made everything way simpler and cleaner to implement.

I totally understand if this is way to much change to accept back upstream.

Besides that I decided to implement a `MavenRepository` class called `TruncatedMavenRepository` (terrible name, i know), which add two new options, such that you can limit the number of .war's and .hpi's pr plugins.  So if you want to keep the last 5 Jenkins versions and 3 plugin versions, that is now possible.  This reduced my download folder from about 100GiB to ~19GiB (as far as I remember).

It still doesn't clean up your local `.m2/repositories/` folder though.  This is a option I would also like to add at some point.  Without this, you need to manually wipe that folder once in a while to still get any space savings.  But hey, at least i'm not storring everything anymore. 

The above is implemented in my [feature/private-update-center](https://github.com/reenberg/update-center2/tree/feature/private-update-center) branch.

I also got a bit fed up with not having propper output.  When syncing from scratch, you will have quite a big period of time with nothing happening on the screen, as it is downloading large ,war files.  I added a Progress Bar as a `TransferListener` which the `wagon` class can take and utilize, in order to make this a bit more user friendly, I also implemented logging using log4j 2 instead of those `System.out` / `System.err` which makes it way better when running unattended for log collection, as it now also has actual debug output.

This needs some more attention, as most of the original `System.out` was turned into INFO logs, but I think at least some of them should get a second opinion.  Also I have just added a very crude `log4j2.xml` which contains class names, etc.  This takes up quite a bit of screen real estate, and is ugly to look at.  This needs to be simplified before it is really useful.

The progress bar and logging is implemented ontop of the other branch in my [feature/better-logging](https://github.com/reenberg/update-center2/tree/feature/better-logging) branch.

Note that I may still rebase those branches at will when I find bugs or find stuff that I forgot, so don't rely on them 100%.

Comments are more than welcome.

LinuxSuRen:
@reenberg Thanks for sharing your ideas here. There're actually some issues with this PR. But I will continue to fix them.

passliang:
please view https://www.oschina.net/news/111266/jenkins-plugin-mirror  to solve plugins download slowly

LinuxSuRen:
> please view https://www.oschina.net/news/111266/jenkins-plugin-mirror to solve plugins download slowly

There's nothing relationship between this repo and mirror. Please visit [here](https://community.jenkins-zh.cn/t/jenkins/26) to get more help. And thanks for your feedback.

daniel-beck:
FWIW this is still on my backlog to review and test in depth; unfortunately I have not yet found the time to do that.

src/main/java/org/jvnet/hudson/update_center/Main.java
```diff
@@ -87,6 +93,15 @@
     @Option(name="-download",usage="Build mirrors.jenkins-ci.org layout")
     public File download = null;
 
+    @Option(name="-cache",usage="Cache target plugin")
```
daniel-beck: This documentation comment makes no sense to me.

This is a directory into which we download from the primary download location, i.e. the cache directory.

```suggestion
    @Option(name="-cache",usage="Cache directory used for caching plugin files")
```

src/main/java/org/jvnet/hudson/update_center/Main.java
```diff
@@ -87,6 +93,15 @@
...
+    public File cache = null;
+
+    @Option(name="-cacheAll",usage="Cache all version of plugin")
```
daniel-beck:
```suggestion
    @Option(name="-cacheAll",usage="Cache all version of plugin instead of only the latest")
```

daniel-beck: I am not sure this has a legitimate use case, as the update center will not serve past releases. Why not e.g. `rsync` or `wget -r` from mirrors.jenkins.io instead to build your archive?

src/main/java/org/jvnet/hudson/update_center/Main.java
```diff
@@ -68,6 +73,7 @@
     public File urlmap = new File("plugin-to-documentation-url.json");
 
     private Map<String, String> pluginToDocumentationUrl = new HashMap<>();
+    private Properties whitelistPro = new Properties();
```
daniel-beck: I've never seen `Pro` used as shorthand for `Properties` so I was pretty confused at first.

Maybe just like this?

```suggestion
    private Properties whitelistProperties = new Properties();
```

(+ adapt all uses)

src/main/java/org/jvnet/hudson/update_center/Main.java
```diff
@@ -158,6 +173,9 @@
     @Option(name="-arguments-file",usage="Specify invocation arguments in a file, with each line being a separate update site build")
     public File argumentsFile;
 
+    @Option(name="-whitelist",usage="White list for plugins")
```
daniel-beck:
```suggestion
    @Option(name="-whitelist",usage="If set to a file using the Java properties format, only plugins listed as keys will be included in the generated update site")
```

README.md
```diff
@@ -110,3 +110,31 @@ you can try to use the appassembler plugin as described below. The exec:java plu
...
+
+Jenkins offical update center has lots of plugins. You could follow below steps to create your own
+private update center:
+
+1. Create your own certificate. For example:
+
+    openssl genrsa -out rootCA/demo.key 1024
+    openssl req -new -x509 -days 1095 -key rootCA/demo.key \
+        -out rootCA/demo.crt \
+        -subj "/C=CN/ST=GD/L=SZ/O=vihoo/OU=dev/CN=demo.com/emailAddress=demo@demo.com"
```
daniel-beck: Please ensure that this code block looks reasonable when rendered as Markdown.

README.md
```diff
@@ -110,3 +110,31 @@ you can try to use the appassembler plugin as described below. The exec:java plu
...
+2. Fetch plugins information then generate update.json
+
+    echo "localization-zh-cn=1.580.1" > whiteList.properties
+    mvn package appassembler:assemble
+	sh target/appassembler/bin/app -id default -www www \
+		-skip-release-history -cache plugins -whitelist whiteList.properties \
+		-key rootCA/demo.key -certificate rootCA/emo.crt \
+		-root-certificate rootCA/demo.crt \
+		-cacheServer http://localhost:9090/plugins/ \
+		-connectionCheckUrl http://localhost:9090/
```
daniel-beck: Please ensure that this code block looks reasonable when rendered as Markdown.

README.md
```diff
@@ -110,3 +110,31 @@ you can try to use the appassembler plugin as described below. The exec:java plu
...
+
+1. Create your own certificate. For example:
+
+    openssl genrsa -out rootCA/demo.key 1024
+    openssl req -new -x509 -days 1095 -key rootCA/demo.key \
+        -out rootCA/demo.crt \
+        -subj "/C=CN/ST=GD/L=SZ/O=vihoo/OU=dev/CN=demo.com/emailAddress=demo@demo.com"
+2. Fetch plugins information then generate update.json
+
+    echo "localization-zh-cn=1.580.1" > whiteList.properties
```
daniel-beck: It is unclear what the property content value is used for, as we only ever check the key exists. This does not appear to select a specific version…?

src/main/java/org/jvnet/hudson/update_center/Main.java
daniel-beck: Nit: The majority of existing options use `-foo-bar` format for parameters, so please use that instead.

src/main/java/org/jvnet/hudson/update_center/Main.java
```diff
@@ -87,6 +93,15 @@
     @Option(name="-download",usage="Build mirrors.jenkins-ci.org layout")
     public File download = null;
 
+    @Option(name="-cache",usage="Cache target plugin")
+    public File cache = null;
+
+    @Option(name="-cacheAll",usage="Cache all version of plugin")
+    public boolean cacheAll;
+
+    @Option(name="-cacheServer",usage="Cache server will replace the orignal")
```
daniel-beck: Nit: The majority of existing options use `-foo-bar` format for parameters, so please use that instead.

LinuxSuRen:
> Requesting that previous feedback is addressed.

I see it. Sorry, I'll do it as soon as possible.

daniel-beck:
@LinuxSuRen No worries, just making sure this doesn't get stalled any longer 👍

LinuxSuRen:
> @LinuxSuRen No worries, just making sure this doesn't get stalled any longer 👍

Sure. Totally understand. And thanks for the reviewing work.

src/main/java/org/jvnet/hudson/update_center/Main.java
LinuxSuRen: Good idea. But one thing we might worth to do this is that keep one technology in the update-center solution might be better. For example, `rsync` or `wget` needs a command line which might dependent on OS platform.

README.md
LinuxSuRen: I might consider do it for matching the properties file format. But the value can actually be ignored.

LinuxSuRen:
@daniel-beck I've finished my changes.

daniel-beck:
@LinuxSuRen Please try https://github.com/jenkins-infra/update-center2/pull/365. It is a big overhaul of the update center generator internally and with new parameters, but after the latest changes, I think everything you need is included:

* `--whitelist-file` is a new parameter for whitelists that are similar to yours. You can specify specific versions, or `*` for all
* The system property or environment variable `DOWNLOADS_ROOT_URL` will be used for download URLs
* `--downloads-directory` (previously `-download`) would write the plugins to a directory tree.

While not exactly the same as what this PR does (especially around "cache"), I think this should take care of your needs. Please let me know how this goes.

daniel-beck:
@LinuxSuRen I just merged #265 which changes a lot about the update center. For interest to you with this PR:

* A whitelist feature exists that seems more flexible than this implementation.
* The whitelist can also basically differentiate between "`-cache`" and "`-cacheAll`" from this PR when running with what used to be the `-download` argument.
* An environment variable or system property named `DOWNLOADS_ROOT_URL` will be used for the download base URL, see the new `HPI.java`.

I think this PR may no longer be needed. Please look at the new version and tell me what you think.

daniel-beck:
As a matter of housekeeping, I'm closing this pull request. This isn't a rejection of the proposed changes; just acknowledging that they're based on top of a much older update-center2 and not usable in their current state (plus they might be obsolete).

I'd still like to know whether the implemented changes I mention in https://github.com/jenkins-infra/update-center2/pull/245#issuecomment-623207864 / https://github.com/jenkins-infra/update-center2/pull/245#issuecomment-633128386 address your use case, please let me know. If not, I'm open to finding a solution together.

LinuxSuRen:
Good to hear more improvements from here. I'll let you know if there're more ideas from me.

PR #246, opened 2018-10-15
remove scriptler from list of ignored plugins

imod:
With version 3.0-alpha (Oct. 10, 2018) - scriptler has fixed all its security vulnerabilities.


imod:
@daniel-beck

daniel-beck:
Alpha releases are not on the regular update center, so this would resume distributing the unsafe releases, see PR build output.

You might be able to achieve the desired effect by removing the entire blacklist by blacklisting each old release individually, similar to `uno-choice`.

PR #247, opened 2018-10-19
[JENKINS-54073] Suppress workflow-job 2.26 for now

jglick:
Pending investigation whether https://github.com/jenkinsci/workflow-api-plugin/pull/81 or something else helps users. CC @svanoort @dwnusbaum; [release notes](https://wiki.jenkins.io/display/JENKINS/Pipeline+Job+Plugin)


jglick:
Well, certainly critical enough for some users to want to roll back. We of course have no way of knowing whether the problem only affects a handful of instances running heavy / weird workloads, or whether there are hundreds of other users suffering in silence. :-/

PR #252, opened 2018-11-22
Docs: Add developer guidelines

oleg-nenashev:
Just minor documentation update towards my experiments with Java 11.

@batmat @alecharp

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ target
 www
 .htaccess
 .idea
+output
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -1,0 +1,10 @@
+Developer Guidelines
+-------
+
+## Quickstart
+
+1. Run `mvn clean package` to build the CLI executable
+2. Run `make test-weekly` to generate the update site in the `output` directory
+  * First run will take a while, 
+    because the update center generator will need to checkout many plugins
+3. Checkout files generated in `output`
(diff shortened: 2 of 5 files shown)


oleg-nenashev:
CC @daniel-beck

batmat:
@oleg-nenashev plan to merge this also?

oleg-nenashev:
@batmat I need to address concerns from @daniel-beck about the Makefile feasibility.
As an alternative, I can detach docs to a separate PR. But then there will be no Quickstart

timja:
> @batmat I need to address concerns from @daniel-beck about the Makefile feasibility.
> As an alternative, I can detach docs to a separate PR. But then there will be no Quickstart

What's the concerns from @daniel-beck ?
This would be a good addition (I'm using this branch to set this up locally)

daniel-beck:
I see no reason to add `make` as a dependency here. It provides no benefits over just straightforward shell scripts.

oleg-nenashev:
@daniel-beck Do you mean the `Jenkinsfile` dependency only? Or do you want to avoid the presence of Make for development purposes as well?

daniel-beck:
@oleg-nenashev More the latter than the former, but ideally both. I do not see what benefits it offers compared to the cost of more dependency.

daniel-beck:
I consider this obsolete after improvements to the documentation here: https://github.com/jenkins-infra/update-center2#invocation

PR #255, opened 2018-12-11
Ignore binary-deployer as deployed by mistake

alecharp:
As mentioned by @daniel-beck in jenkins-infra/repository-permissions-updater#943, filling this PR to ignore `binary-deployer`. The project was / is a POC and not meant to be deployed. It's not hosted in `jenkinsci` org either.


daniel-beck:
You have users!
https://stats.jenkins.io/plugin-installation-trend/binary-deployer.stats.json

alecharp:
what?! no, never was supposed to have any.. plugin is great *cough* *cough* but shouldn't be used.. Thanks anyway @daniel-beck

PR #256, opened 2018-12-20
[JENKINS-55159] Add hpaat plugin 5.6 to the blacklist

grount:
The name of the plugin "hp-application-automation-tools-plugin" should be written as in the following link?
[repo.jenkins-ci.org](https://repo.jenkins-ci.org/releases/org/jenkins-ci/plugins/hp-application-automation-tools-plugin/)

Also indented the comments after #.
Let me know if any changes are required.

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -142,18 +142,19 @@ kubernetes-pipeline-devops-steps-1.3
 kubernetes-pipeline-aggregator-1.3
 
 # specific releases
-active-directory-2.1      # JENKINS-42739
-BlameSubversion-1.121     # big number but not the latest release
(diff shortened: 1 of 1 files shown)


daniel-beck:
@grount Sorry I missed the notification for this. Should we still merge this?

grount:
@daniel-beck Yeah it still should be blacklisted, I've released 5.6.2 instead.

daniel-beck:
@grount Thanks! Since both share the same core dependency, 5.6.2 will always supersede 5.6, so this isn't needed.

grount:
@daniel-beck my fear is that It is still installable if not banned is it?

daniel-beck:
Good point. It is currently still listed in the archives.

PR #258, opened 2019-02-07
Hide git-client 3.0.0-rc and git 4.0.0-rc

batmat:
As requested by @markewaite

The expectation was to land these releases into the experimental UC only.

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -304,3 +304,8 @@ config-file-provider-3.0
 
 # JENKINS-54073, claimed performance regression as of yet undiagnosed; plus some more minor issues:
 workflow-job-2.26
+
+# Requested by Mark Waite https://groups.google.com/d/msg/jenkinsci-dev/9MDl1SWeF0o/cvp3_nzjFAAJ
+# mistakenly used -rc suffix thinking it would only be exposed in the experimental update center
+git-4.0.0-rc
+git-client-3.0.0-rc


batmat:
I guess we wait for Mark feedback a bit, and we merge if he cannot check it soon.
I checked the git and git-client plugins repos, and AFAICT I should have put everything with -rc suffix.

Worst case if we merge before Mark can approve, we file another one or adjust. But this will hide these releases quicker.

daniel-beck:
This exactly matches Mark's request in the mailing list, so I see no reason to wait.

MarkEWaite:
Thanks very much for doing this!

My sincere apologies that I didn't read the experimental update center instructions before choosing this version string.

CptPlastic:
UGH this change broke our jenkins servers we rolled back to test something and now everything depends on 3.0.0.rc to bet there.

MarkEWaite:
@CptPlastic there should be nothing which depends on 3.0.0-rc except the git plugin.  If you rollback the git plugin to git plugin 3.9.3 and the git client plugin to 2.7.4 it should resolve all dependencies successfully.  No other plugin that I've detected has a dependency on git client plugin 3.0.0-rc.

CptPlastic:
This fixed the pipelines but stash is jacked. 
image

Also the downgrade paths seems messed up. This should have just been fixed in another release removing the 3.0.0.rc from the update center caused issues. I get it that having them in there is not idea but it could have been fixed with an upstream update then removed after some time. What a pain.

MarkEWaite:
No @CptPlastic it would not have been better to fix it in another release that superseded 3.0.0-rc.  That new release would need to have included only the code from git client plugin 2.7.6 (without any of the 3.0.0 capabilities) and would have created much more confusion than the removal from the update center.  

My mistake with 3.0.0-rc was already quite severe in the community, as evidenced by the 5+ bug reports that appeared within the first week of that release.  I don't want to compound the mistake by delivering something named "3.0.0-*" that does not include the 3.0.0 functionality.

PR #260, opened 2019-02-22
Remove the obsolete fortify360 plugin

akaryakina:
We would like to remove or deprecate the old Fortify360 plugin. It's obsolete and has not worked for years. It's been completely replaced with the Fortify plugin that's been open sourced a few days ago [https://github.com/jenkinsci/fortify-plugin](https://github.com/jenkinsci/fortify-plugin)
Fortify360 is the name that's not used by our product management anymore, and having this plugin could be very confusing for our customers. The new plugin's wiki is here [https://wiki.jenkins.io/display/JENKINS/Fortify+Plugin](https://wiki.jenkins.io/display/JENKINS/Fortify+Plugin)
Also, here's the relevant JIRA ticket with more details: [HOSTING-710](https://issues.jenkins-ci.org/browse/HOSTING-710)

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -309,3 +309,6 @@ workflow-job-2.26
 # mistakenly used -rc suffix thinking it would only be exposed in the experimental update center
 git-4.0.0-rc
 git-client-3.0.0-rc
+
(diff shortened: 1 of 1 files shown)


daniel-beck:
The plugin is ownerless per uploader permissions. Requester is the only GitHub user with commit access. So this looks good authorization-wise.

PR #261, opened 2019-02-26

akamen:
Removing blackduck-hub


daniel-beck:
Requester is in https://github.com/orgs/jenkinsci/teams/blackduck-hub-plugin-developers

PR #262, opened 2019-03-01
No further releases of 1.3.x should match via the 1.3 clause

daniel-beck:


src/main/resources/warnings.json
```diff
@@ -3182,7 +3182,7 @@
     "versions": [
       {
         "lastVersion": "1.3.4",
-        "pattern": "(?!1[.]3[.]4[.].*)(0|1[.]([0-2]|3|3[.][1-4]))(|[.-].*)"
+        "pattern": "(?!1[.]3[.]4[.].*)(0|1[.]([0-2]|3$|3[.][1-4]))(|[.-].*)"
```
abayer: This would match `1.3.10` - I just came up with `(?!1[.]3[.]([3-4][.].*|[5-9].*|\d\d.*))(0|1[.]([0-2]|3|3[.][1-4]))(|[.-].*)` which is ugly but does work.

abayer: Whoops, I mistested. Is good.

PR #268, opened 2019-04-11
Remove bad release for allure-jenkins-plugin

eroshenkoam:


src/main/resources/artifact-ignores.properties
```diff
@@ -301,6 +301,9 @@ reviewboard         # depends on perforce
 websphere-deployer-1.5.5
 websphere-deployer-1.5.6
 
+# requested by maintenier
```
daniel-beck:
```suggestion
# requested by maintainer
```

daniel-beck:
@eroshenkoam Could you provide a reference to a bug report or similar for posterity?

eroshenkoam:
Last version of allure-jenkins-plugin is 2.28.1
https://github.com/jenkinsci/allure-plugin/releases

I mixed up the release version (3.11.0) and I worry that users will be confused

PR #277, opened 2019-05-28
Swarm 3.16 fixed SECURITY-1252

daniel-beck:
@basil


daniel-beck:
Given https://github.com/jenkinsci/swarm-plugin/pull/90 it might make more sense to declare it fixed in 3.17?

basil:
Agreed, let's declare 3.17 as the first fixed release to avoid any ambiguity.

daniel-beck:
To be merged once PR build passed.

PR #281, opened 2019-06-13
[INFRA-2151] Fixed redirect issues with some files

mbarbero:
Redirect issues with
* update-center.actual.json, 
* latestCore.txt, and 
* plugin-documentation-urls.json

Tests have been improved to show the issue and avoid any later regression


daniel-beck:
(Obligatory note that this is all internal implementation details and should only be considered supported as far as actual Jenkins deliverables rely on it.)

daniel-beck:
@mbarbero Thanks again for this! 👍

mbarbero:
@daniel-beck Anytime. Cheers.

zbynek:
@mbarbero this breaks `https://updates.jenkins.io/latestCore.txt` for Jenkins weekly. Note that each rewrite condition only applies to the next rule. @daniel-beck looped

daniel-beck:
Right, I missed (like @mbarbero I guess) that all `RewriteCond` only apply to a single `RewriteRule`, rather than all following.

PR #282, opened 2019-06-15
deprecate configuration-as-code-support

casz:
configuration-as-code-support is no more!


daniel-beck:
Just in case, did you consider the experience of users not yet on the tombstone release?

casz:
@daniel-beck 1.18 was marked incompatible so if they do see that in their update center it will show up as incompatible.

Anyone having an old version of support plugin does not matter.

Either 
The instance will be spun up become unusable prompting them to go look either at releases page on JCasC or start googling or create an issue and get their answer... Update your plugins :sweat: 

or everything will continue to work. Since these are just "support" configurators and those support configurators that were moved have a higher ordinal meaning they will load first.

Not too concerned.

PR #292, opened 2019-08-07
Stop distributing gcm-notification as the GCM service is dead

orrc:
The Google Cloud Messaging (GCM) service no longer exists, so presumably the plugin no longer works, and hasn't been maintained in years anyway.


MarkEWaite:
A version of this change has already been applied to master branch. We can safely close this one.  Thanks @orrc and @daniel-beck !

daniel-beck:
https://github.com/jenkins-infra/update-center2/blob/87c8df7be5c116fe95f6f52a6cad02c32e091cda/src/main/resources/artifact-ignores.properties#L341

PR #293, opened 2019-08-09
mark qtest-1.4.3 as invalid release

phuthanhta:
please mark qtest-1.4.3 as invalid release. I missed one step in release progress.


daniel-beck:
A user with commit access to https://github.com/jenkinsci/qtest-plugin needs to confirm this request.

PR #299, opened 2019-09-16
Remove codescan - 

ustramooner:
the service has been shut down and replaced with an alternate system.

following instructions from here
https://jenkins.io/doc/developer/publishing/removing-from-distribution/

please indicate if additional steps required


daniel-beck:
This doesn't accomplish what you expect. Follow the instructions in https://github.com/jenkins-infra/update-center2#removing-plugins-from-distribution

ustramooner:
> This doesn't accomplish what you expect. Follow the instructions in https://github.com/jenkins-infra/update-center2#removing-plugins-from-distribution

is that better?

daniel-beck:
@ustramooner Looks good, thanks!

PR #306, opened 2019-10-30
[SECURITY-1433] Limit affected versions

timsotack:
This update is to indicate that SECURITY-1433 has been fixed in 4.0.1 version of the plugin.


daniel-beck:
Regex LGTM. It also matches exactly the versions it should per test output from the PR build.

I plan to review the actual fix tomorrow, and then merge this if the fix addresses the problem.

daniel-beck:
Due to infra problems I'm currently unable to set up a test instance to confirm it's resolved.

timsotack:
Round trip from browser to jenkins server?

timsotack:
The original issue is credentials stored in plaintext which has been resolved. What you have highlighted is a new, separate, lower severity issue. I think the proper course would be to validate the first issue is fixed and log a separate vuln against 4.0.1 which i am happy to fix once I completey understand the issue.

daniel-beck:
Filed https://github.com/jenkinsci/fortify-on-demand-uploader-plugin/pull/85 to address the leftovers.

Meanwhile I consider this to be an incomplete fix and will not treat it as an independent issue. Our security advisories are hours of work from multiple people, and involves notifying thousands of users, so I'm not going to do that just to downgrade this issue from plaintext storage+transmission to plaintext transmission.

daniel-beck:
Additionally,

> These credentials can be viewed by users with Extended Read permission ….

This still works through the config UI.

> Screenshot 2019-11-05 at 21 38 33

This user only has Extended Read permission (no Save/Apply buttons), and can obtain the password from the form field.

timsotack:
Just to be clear will https://github.com/jenkinsci/fortify-on-demand-uploader-plugin/pull/85 address the Extended Read permission issue cited above?

daniel-beck:
@timsotack Yes. That's basically the only thing I tested. It's trivial to see when testing too, since the encrypted value is always super long afterwards.

timsotack:
The issues is completely addressed in 5.0.0 which will be released early December. I will create a new PR then to modify the affected version regex .

PR #309, opened 2019-11-05
Remove pipeline-model-definition-1.5.0-rc1

bitwiseman:


MarkEWaite:
Aren't there 4 plugins that need to be blacklisted, not just one?
image

PR #318, opened 2019-11-25
Reintroduce versioning of the update center

oleg-nenashev:
Update center does not longer require Maven releases since 2.0 by @daniel-beck . But it would be great to have some kind of versioning to have a history of changes and to highlight contributions.

My suggestion:

* We switch to Semver
* We start from a 3.0.0 version or 2.1.0 version
* We use a `v1.2.3` format recommended by GitHub

LKater we can switch to a GitOps mode for the repository to simplify management of the deployments

CC @halkeye @olblak for their opinions

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,6 @@
+# See https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.adoc
+_extends: jenkinsci/.github
+# Semantic versioning: https://semver.org/
+version-template: $MAJOR.$MINOR.$PATCH
+tag-template: v$NEXT_MINOR_VERSION
+name-template: v$NEXT_MINOR_VERSION


daniel-beck:
> Update center does not longer require Maven releases since 2.0 by @daniel-beck

Not true, it didn't really have versions for many years before then. The only problem was that the tools crawler uses update-center2 as a dependency, so needs to be roughly aligned with some basic behaviors, necessitating "releases" that are actually unused.

oleg-nenashev:
Thanks for the clarification! I will try to document it in README

daniel-beck:
https://github.com/jenkins-infra/crawler/blob/27620816cf86727eb9638155e8a1d1392c817d23/pom.xml#L35-L37

https://github.com/jenkins-infra/backend-extension-indexer/blob/aab1f59f878ac8a28ad17c235f8b99b52447b233/pom.xml#L58-L60

(And extension indexer is also used by https://github.com/jenkins-infra/pipeline-steps-doc-generator/blob/master/Jenkinsfile#L21-L39)

daniel-beck:
> roughly aligned with some basic behaviors

IIRC it specifically was the signing and how long before cert expiration the build starts failing. KK's general unavailability necessitates some date juggling.

daniel-beck:
I do not think this change is justified. The stated purpose is unrelated to technical versioning necessities, and only aimed at supporting the tooling established elsewhere. The history of changes, and authors of contributions, can easily be determined through the git log and the "Insights".

daniel-beck:
This seems obsolete. Feel free to reopen if it's not.

PR #329, opened 2020-01-29
Do not match fixed LTS version in the weekly versions regex.

daniel-beck:
From https://github.com/jenkins-infra/update-center2/commit/3832d80090909734924f13b179ec10f454ce782a#r37023469

Generating this regex wasn't the best idea after all…


daniel-beck:
> It's even smaller. Nice!

Unfortunately, because it's no longer automatically generated 😦

PR #316, opened 2019-11-22
INFRA-2240 - If scm tag contains github, then try to pull topics from it

halkeye:
https://issues.jenkins-ci.org/browse/INFRA-2240

Potentially crappy way to do it, one github api call per plugin.

PR #331, opened 2020-02-02
Expose `issueTracker`

timja:
Ref:
https://groups.google.com/d/msg/jenkinsci-dev/haFTYlhp7h8/hkApal9WAgAJ

Uses the github api to check if GitHub issues are enabled, else sets
Jenkins JIRA, if there's any error it will fallback to Jenkins JIRA as
well.

Samples:

```
      "issueTracker": {
        "name": "Jenkins Jira",
        "url": "https://issues.jenkins-ci.org/"
      },
```
```
      "issueTracker": {
        "name": "GitHub Issues",
        "url": "https://github.com/jenkinsci/configuration-as-code-plugin/issues"
      },
```

For testing I patched:
`TruncatedMavenRepository` locally in `listHudsonPlugins`
with:
```java
        PluginHistory pluginHistory = base.listHudsonPlugins()
                .stream()
                .filter(plugin -> StringUtils.equals(plugin.artifactId, "configuration-as-code"))
                .findAny()
                .orElseThrow(IllegalArgumentException::new);

        return Collections.singletonList(pluginHistory);
```
and for jira:
```java
                .filter(plugin -> StringUtils.equals(plugin.artifactId, "mailer"))
```

Ran with:

```shell
mvn package appassembler:assemble -DskipTests
sh target/appassembler/bin/app -id com.example.jenkins -www www -maxPlugins 1
```

### DIFF
--- a/src/main/java/org/jvnet/hudson/update_center/Plugin.java
+++ b/src/main/java/org/jvnet/hudson/update_center/Plugin.java
@@ -25,6 +25,7 @@ package org.jvnet.hudson.update_center;
 
 import com.google.common.annotations.VisibleForTesting;
 import hudson.util.VersionNumber;
+import javax.annotation.CheckForNull;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import org.apache.commons.io.IOUtils;
@@ -35,6 +36,7 @@ import org.dom4j.DocumentFactory;
 import org.dom4j.Element;
 import org.dom4j.Node;
 import org.dom4j.io.SAXReader;
+import org.jvnet.hudson.update_center.util.UrlToGitHubSlugConverter;
 import org.sonatype.nexus.index.ArtifactInfo;
 
 import java.io.File;
@@ -81,8 +83,16 @@ public class Plugin {
      * Previous version of this plugin.
      */
     public final HPI previous;
-    
+
     private final SAXReader xmlReader;
+
+    private static final JSONObject JIRA_ISSUE_TRACKER;
+
+    static {
+        JIRA_ISSUE_TRACKER = new JSONObject();
+        JIRA_ISSUE_TRACKER.put("name", "Jenkins JIRA");
+        JIRA_ISSUE_TRACKER.put("url", "https://issues.jenkins-ci.org/");
+    }
 
     /**
      * POM parsed as a DOM.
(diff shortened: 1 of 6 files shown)


halkeye:
So why not do this based on issue tracker in the pom? 

Also take a look at #316 where I used the graphql API to prevent the need to do 1700 GitHub calls.

timja:
> So why not do this based on issue tracker in the pom?
> 
> Also take a look at #316 where I used the graphql API to prevent the need to do 1700 GitHub calls.

Because it's not there already and it would take a lot of effort to get people to add it, release their plugin etc, this way we just detect it.

src/main/java/org/jvnet/hudson/update_center/Plugin.java
```diff
@@ -81,9 +83,17 @@
...
     public final HPI previous;
-    
+
     private final SAXReader xmlReader;
 
+    private static final JSONObject JIRA_ISSUE_TRACKER;
+
+    static {
+        JIRA_ISSUE_TRACKER = new JSONObject();
+        JIRA_ISSUE_TRACKER.put("name", "Jenkins JIRA");
```
oleg-nenashev:
```suggestion
        JIRA_ISSUE_TRACKER.put("name", "Jenkins Jira");
```

src/main/java/org/jvnet/hudson/update_center/Plugin.java
```diff
@@ -357,6 +367,32 @@ private String requireGitHubRepoExistence(String url) {
...
+                boolean issuesEnabled = GitHubSource.getInstance().issuesEnabled(slug);
+
+                if (issuesEnabled) {
+                    return githubIssueTracker("https://github.com/" + slug + "/issues");
+                }
+            } catch (Exception e) {
+                System.out.println("Failed to check if issues are enabled: " + e.getMessage());
+            }
+        } else {
+            System.out.println("SCM url is empty, defaulting to Jenkins JIRA for issue tracking");
```
oleg-nenashev:
```suggestion
            System.out.println("SCM url is empty, defaulting to Jenkins Jira for issue tracking");
```

halkeye:
I really strongly think this should be converted to graphql and pre fetched like the labels pr. This is going to do more than 1700 API calls and really slow things down.

Also the CI (testing not building) runs things without a token enabled, so it's limited to anon request speed

timja:
There’s already an api request and I assumed this would be in the cache for it? Haven’t checked closely that the cache is hit though

src/main/java/org/jvnet/hudson/update_center/util/UrlToGitHubSlugConverter.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private UrlToGitHubSlugConverter() {
+    }
+
+    /**
+     * Converts a github url to its 'slug'
+     * @param url a url like: https://github.com/jenkinsci/blueocean-plugin
+     * @return the github slug, jenkinsci/blueocean-plugin
+     */
+    public static String convert(@Nonnull String url) {
```
daniel-beck: Do we expect strings like `http://github.com/org/repo`, `https:/github.com/org/repo`, `https://github.com/org/repo/repo`, `git@github.com:org/repo.git`, etc. to get here?

daniel-beck:
Looking at the GitHub API, this seems to result in an additional 1700 requests every four minutes (except that it'd take a lot more than 4 minutes). While it looks like it wouldn't exhaust our API rate limit based on https://github.com/jenkins-infra/update-center2/blob/7adc44b51ef9ada205a517917ceeabba9148c635/src/main/java/org/jvnet/hudson/update_center/GitHubSource.java#L30 it would probably still take FOREVER.

daniel-beck:
Those ~22 requests listing the repos every hour take a minute, by the way.

> 01:26:07 ** Falling back to default repo for accelerated-build-now-plugin: https://github.com/jenkinsci/accelerated-build-now-plugin-plugin
> 01:26:07 Retrieving GitHub repository names...
> 01:27:04 => org.terracotta.jenkins.plugins:accelerated-build-now-plugin:1.0.1

timja:
I’ll take a look at converting it to graphq once halkeye’s pr is merged

src/main/java/org/jvnet/hudson/update_center/util/UrlToGitHubSlugConverter.java
timja: From what I remember it’s using the already resolved repo name from earlier but will check when I look at adapting this to graphql

daniel-beck:
Superseded by https://github.com/jenkins-infra/update-center2/pull/355

PR #332, opened 2020-02-05
[SECURITY-1177] Limit affected versions

guidograzioli:
Limit security warning to [0.1,0.13.*], 0.14 released with fix:
https://github.com/jenkinsci/depgraph-view-plugin/pull/18

Advisory: 
https://jenkins.io/security/advisory/2019-07-11/#SECURITY-1177

### DIFF
--- a/src/main/resources/warnings.json
+++ b/src/main/resources/warnings.json
@@ -5058,7 +5058,7 @@
     "url": "https://jenkins.io/security/advisory/2019-07-11/#SECURITY-1177",
     "versions": [
       {
-        "lastVersion": "0.14",
+        "lastVersion": "0.13",
         "pattern": "0[.]([1-9]|1[0-3])(|[.-].*)"
       }
     ]
--- a/src/main/resources/warnings.json
+++ b/src/main/resources/warnings.json
@@ -5058,8 +5058,8 @@
     "url": "https://jenkins.io/security/advisory/2019-07-11/#SECURITY-1177",
     "versions": [
       {
-        "lastVersion": "0.13",
-        "pattern": ".*"
+        "lastVersion": "0.14",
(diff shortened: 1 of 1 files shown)


src/main/resources/warnings.json
```diff
@@ -5058,8 +5058,8 @@
     "url": "https://jenkins.io/security/advisory/2019-07-11/#SECURITY-1177",
     "versions": [
       {
-        "lastVersion": "0.13",
-        "pattern": ".*"
+        "lastVersion": "0.14",
```
daniel-beck:
```suggestion
        "lastVersion": "0.13",
```

Last version is the last _affected_ version, for messages like "affected versions up to and including X".

daniel-beck:
Confirmed fixed in the current release:

> Screenshot

daniel-beck:
Jenkins security team internal note: Our tooling generates the following regex for this:

> `(0[.]1[0-3]|0[.][2-9])(|[.-].+)`

guidograzioli:
> Jenkins security team internal note: Our tooling generates the following regex for this:
> 
> > `(0[.]1[0-3]|0[.][2-9])(|[.-].+)`

Ah! Wasn't aware there was a tool, so i built it manually on regexplanent. Would you like me to update this as well? (question: why does it not match 0.1 ?)

src/main/resources/warnings.json
guidograzioli: Applied

daniel-beck:
> Wasn't aware there was a tool

There isn't (internal use only). Sorry for the confusion.

> Would you like me to update this as well?

Nope, yours looks good too! We don't discriminate against hand-crafted regexes 😁

> (question: why does it not match 0.1 ?)

It doesn't match 0.1 because that release doesn't exist.

PR #334, opened 2020-02-07
Add labels for macstadium-orka

ispasov:


daniel-beck:
https://github.com/jenkinsci/macstadium-orka-plugin

PR #338, opened 2020-02-12
Future versions have a fix for this problem

MidVision:
Hi team, a simple update for versioning containing a fix for this particular problem.


src/main/resources/warnings.json
```diff
@@ -6830,7 +6830,7 @@
     "versions": [
       {
         "lastVersion": "4.1",
-        "pattern": ".*"
+        "pattern": "([123][.].+|4[.][0-1])(|[.-].*)"
```
daniel-beck: Can be simplified further, but looks fine 😃 

```suggestion
        "pattern": "([123]|4[.][01])(|[.-].*)"
```

MidVision: Thanks for your recommendation Daniel! Needless to say, feel free to use it instead of mine! :-)

PR #339, opened 2020-02-17
[SECURITY-727] Second half addressed in 1.14.0

daniel-beck:
According to maintainers, this has been addressed in release 1.14.0.

I need to confirm this, then this can be merged.


daniel-beck:
Confirmed fixed.

PR #340, opened 2020-02-19
[INFRA-2478] - override the Wiki URL by the GitHub URL for qtest plugin

tamvanvo:


daniel-beck:
The plugin is actively maintained and just needs a new release for its recent POM change to be effective.

Because of that I'd rather not use this override and introduce what amounts to tech debt.

tamvanvo:
I have the jira ticket to explain this pull request https://issues.jenkins-ci.org/browse/INFRA-2478

daniel-beck:
My previous comment is nonblocking, go ahead 🤷‍♂

PR #342, opened 2020-02-19
Remove the adopt-me label from the list of whtelisted ones, use adopt-this-plugin

oleg-nenashev:
Aligns with the current adoption process: https://github.com/jenkins-infra/jenkins.io/pull/2875


oleg-nenashev:
CC @uhafner , it should address your comment in the dev list

PR #343, opened 2020-02-23
Make GitHubSource work without API credentials

daniel-beck:
The idea is to have basic operation without credentials anywhere, this was broken in https://github.com/jenkins-infra/update-center2/pull/316.

Also reformat the test file, it was CRLF formatted.

Also, use `org/repo` format for the collection; we were in a weird in-between state where the org name is pluggable on topic collection, but the data store was not. Improve that a bit.


src/test/java/org/jvnet/hudson/update_center/GitHubSourceTest.java
```diff
@@ -1,46 +1,54 @@
...
+            }
+
+            @Override
+            public List<String> getRepositoryTopics(String org, String repo) throws IOException {
+                return getOrganizationTopics(org).getOrDefault(org + "/" + repo, Collections.emptyList());
+            }
+        };
+        assertEquals(
+                Arrays.asList("pipeline"),
+                gh.getRepositoryTopics("jenkinsci", "workflow-cps-plugin")
```
daniel-beck: Order of expected and actual was wrong here.

daniel-beck:
@halkeye PTAL

halkeye:
I mean it looks good to me at a glance. I'll try it out locally when I'm next online.

What was the actual test to support without credentials? I thought it was working cause CI (not trusted/release) did it without credentials

daniel-beck:
> What was the actual test to support without credentials?

This tool isn't well supported by autotests.

> I thought it was working cause CI (not trusted/release) did it without credentials

It doesn't abort, just prints errors.

halkeye:
> This tool isn't well supported by autotests.

sorry, i meant what was the change, I couldn't figure it out from the diffs

daniel-beck:
> what was the change

`#getRepositoryTopics` (formerly two-arg `#getTopics`) no longer calls `#getOrganizationTopics` (formerly one-arg `#getTopics`) to potentially query Github; instead, we assume there are no topics if they weren't done during `#init` (and that has a condition checking whether credentials are set).

daniel-beck:
Closed in https://github.com/daniel-beck/update-center2/commit/ce0b447d038aaaee7b95ec5eabb7f0f459aa5dee

PR #344, opened 2020-02-23
[JENKINS-59136] Deprecation support

daniel-beck:
Companion PR to https://github.com/jenkinsci/jenkins/pull/4073 for a basic implementation of [JENKINS-59136](https://issues.jenkins-ci.org/browse/JENKINS-59136).

---

Add support for marking any plugin artifact ID as deprecated, even those not currently published.

Broadly speaking, there are now two supported mechanisms for marking as deprecated:

* The label `deprecated`.
* Entries in a new top-level object `deprecations`, keys are plugin IDs, and values are objects with a key `url` whose value is a String.

The former requires no changes to the `update-center.json` format. This implements the latter.

There are two use cases for the top-level `deprecations` element:

* Deprecating plugins that are no longer published.
* Using a different deprecation reference URL than just the regular plugin URL (on plugins.jenkins.io). URLs can be mailing list archives, Jira issues, PR discussions, whatever makes sense. 

A new resource file `deprecations.properties` can be used to define deprecations for the top-level element.

For now this implementation does not add "duplicate" entries to the top-level `deprecations` map from regular labels, but that could be added in the future for easier consumption. We probably need to continue defining the regular `deprecated` label for backwards compatibility.

Sample data:

```
"deprecations": {
  "i-am-an-invalid-plugin-id": {
    "url":"https://github.com/jenkins-infra/update-center2"
  }
}
```

This format allows for some future extensibility at a fairly low complexity cost (as opposed to the minimal `{ "pluginId": "url" }`).

### DIFF
--- a/resources/deprecations.properties
+++ b/resources/deprecations.properties
@@ -3,73 +3,134 @@
 # This file allows marking plugins as deprecated while not publishing them otherwise.
 # Additionally, this file allows referencing deprecation notices outside of (obsolete or non-existing) plugin documentation.
 
-appthwack = https://wiki.jenkins.io/display/JENKINS/AppThwack+Plugin
+# https://wiki.jenkins.io/display/JENKINS/AppThwack+Plugin
+appthwack = https://wiki.jenkins.io/x/cwchB
 assembla-oauth = https://groups.google.com/d/msg/jenkinsci-dev/TVz-D5etsUM/Knx9zXtEnSwJ
-blackduck-hub = https://github.com/jenkins-infra/update-center2/pull/261
-blackduck-installer = https://wiki.jenkins-ci.org/display/JENKINS/Black+Duck+Vulnerability+Installer+Plugin
+# https://github.com/jenkins-infra/update-center2/pull/261
+blackduck-hub = https://git.io/JfaQa
+# https://wiki.jenkins-ci.org/display/JENKINS/Black+Duck+Vulnerability+Installer+Plugin
+blackduck-installer = https://wiki.jenkins.io/x/nYHHB
 build-flow-extensions-plugin = https://groups.google.com/d/msg/jenkinsci-dev/YKfydxnpvyE/mMN7LNBoBgAJ
 build-flow-test-aggregator = https://groups.google.com/d/msg/jenkinsci-dev/YKfydxnpvyE/mMN7LNBoBgAJ
 build-flow-toolbox-plugin = https://groups.google.com/d/msg/jenkinsci-dev/YKfydxnpvyE/mMN7LNBoBgAJ
-build-node-column = https://wiki.jenkins-ci.org/display/JENKINS/Build+Node+Column+Plugin
-buildcoin-plugin = https://wiki.jenkins-ci.org/display/JENKINS/Buildcoin+Plugin
-buildheroes = https://wiki.jenkins-ci.org/display/JENKINS/Buildheroes
-caroline = https://wiki.jenkins-ci.org/display/JENKINS/Caroline+Plugin
+# https://wiki.jenkins-ci.org/display/JENKINS/Build+Node+Column+Plugin
+build-node-column = https://wiki.jenkins.io/x/1QiMAw
+# https://wiki.jenkins-ci.org/display/JENKINS/Buildcoin+Plugin
+buildcoin-plugin = https://wiki.jenkins.io/x/KoyhAw
+# https://wiki.jenkins-ci.org/display/JENKINS/Buildheroes
+buildheroes = https://wiki.jenkins.io/x/_AD8Aw
+# https://wiki.jenkins-ci.org/display/JENKINS/Caroline+Plugin
+caroline = https://wiki.jenkins.io/x/AwOMAw
 chrome-frame-plugin = https://www.chromium.org/developers/how-tos/chrome-frame-getting-started
-cifs = https://wiki.jenkins-ci.org/display/JENKINS/CIFS-Publisher+Plugin
-clang-scanbuild-plugin = https://github.com/jenkinsci/clang-scanbuild-plugin/commit/a6d57b67f6fbd0a9893ecf6436c54ecb670d5829
-cloudbees-deployer-plugin = https://wiki.jenkins-ci.org/display/JENKINS/Cloudbees+Deployer+Plugin
-cloudbees-disk-usage-simple-plugin = https://github.com/jenkins-infra/update-center2/pull/179
-cloudbees-enterprise-plugins = https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Jenkins+Enterprise
-cloudbees-registration = https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Registration+Plugin
-codescan = https://github.com/jenkins-infra/update-center2/pull/299
-copyarchiver = https://wiki.jenkins-ci.org/display/JENKINS/CopyArchiver+Plugin
-cppunit = https://wiki.jenkins-ci.org/display/JENKINS/CppUnit+Plugin
-dockerhub = https://github.com/jenkins-infra/update-center2/commit/e5a138873113d0e2d6af59d5699215a1835c15b7
-emotional-hudson = https://wiki.jenkins-ci.org/display/JENKINS/Emotional+Hudson+Plugin
-external-scheduler = https://github.com/jenkins-infra/update-center2/pull/257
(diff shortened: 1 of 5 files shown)


oleg-nenashev:
Should we also add a 'depublished' list? Out of the scope of this PR, but might help in the future

daniel-beck:
> Should we also add a 'depublished' list? Out of the scope of this PR, but might help in the future

Easy enough to add once we need it (assuming we don't want reference URLs, depends on the use case). Meanwhile it'll just be unnecessary traffic.

daniel-beck:
> Should we also add a 'depublished' list? Out of the scope of this PR, but might help in the future

FWIW it's safe to say that a depublished plugin is in _some way_ deprecated; and the way I designed this allows deprecation of depublished plugins. IOW, this basically exists, modulo some necessary data duplication (needs entries in two files).

zbynek:
Would it make sense to extend a format to
```
{
DEPRECATED_ID: {url: URL, replacement: REPLACEMENT_ID}
}
```
so that Jenkins Plugin Manager could offer a "Replace with <plugin>" button?

daniel-beck:
@zbynek There's a start of a related conversation in https://github.com/jenkinsci/jenkins/pull/4073 (and its related Jira issue). Deprecations often might not have a replacement, and there's other potential situations that need fixing. We either have a super elaborate format, or are happy enough with a URL that explains how to proceed.

I also propose a slightly different format to get started with https://github.com/jenkinsci/jenkins/pull/4073#issuecomment-590589651 that would be forward (backward?) compatible with format extensions.

oleg-nenashev:
> We either have a super elaborate format, or are happy enough with a URL that explains how to proceed.

FTR https://jenkins.io/doc/developer/plugin-governance/deprecating-or-removing-plugin/ currently assumes that the reasons are explained in the plugin documentation. We could definitely add more complicated formats, but t requires more discussion.

* Use-case for replacement links are the plugins like https://plugins.jenkins.io/pmd/ . They were marked as deprecated as a part of the #345 processing. If we inject a pointer to a replacement to the warning boxes, it would be great
* Deprecated JS Libs plugins from #346 are a use-case where there is no replacement... and when users actually cannot do anything about deprecations. Maybe deprecation of API plugins should be handled separately somehow

README.md
```diff
@@ -34,14 +34,15 @@ Extra features
...
 
-Only labels in the [whitelist file](./src/main/resources/allowed-labels.properties) will make it into the jenkins infrastructure.
+See https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/model/Messages.properties (look for `UpdateCenter.PluginCategory`) for the localization overrides applied to labels by Jenkins.
+Other labels are categorized into general _Misc (custom-label-here)_ categories.
 
-See https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/model/Messages.properties (look for `UpdateCenter.PluginCategory`) for the localization overrides applied by Jenkins.
-
-Older plugins may have additional labels defined in the file [`label-definitions.properties`](https://github.com/jenkins-infra/update-center2/edit/master/src/main/resources/label-definitions.properties) in this repository, this approach is deprecated.
+Some plugins may have additional labels defined in the file [`label-definitions.properties`](https://github.com/jenkins-infra/update-center2/edit/master/src/main/resources/label-definitions.properties) in this repository, this approach is deprecated.
```
oleg-nenashev: Not really deprecated. This is the only way to set different labels for plugins in multi-module (multi-plugin) repositories

src/main/java/org/jvnet/hudson/update_center/Main.java
```diff
@@ -333,6 +334,28 @@ private JSONObject buildUpdateCenterJson(MavenRepository repo, LatestLinkBuilder
...
+
+            final JSONArray labels = plugin.getJSONArray("labels");
+            if (labels.contains("deprecated")) {
+                labeledPluginsAndURL.put(pluginName, plugin.getString("wiki"));
+            }
+        }
+
+        // Process deprecations via file
+        for (String plugin : Plugin.getDeprecatedPlugins()) {
+            labeledPluginsAndURL.put(plugin, Plugin.getCustomDeprecationUri(plugin));
```
Jetersen: Nice addition.

daniel-beck:
Rewrote the PR on top of https://github.com/jenkins-infra/update-center2/pull/365.

daniel-beck:
Added an initial list of some plugins on the ignore list with at least somewhat useful references.

daniel-beck:
@halkeye @jetersen @oleg-nenashev @timja I merged the overhaul PR that this was based on. Could I get re-reviews, especially wrt to the initial list of deprecations?

jetersen:
Configuration as Code support plugin is missing 😅

resources/deprecations.properties
```diff
@@ -0,0 +1,75 @@
...
+caroline = https://wiki.jenkins-ci.org/display/JENKINS/Caroline+Plugin
+chrome-frame-plugin = https://www.chromium.org/developers/how-tos/chrome-frame-getting-started
+cifs = https://wiki.jenkins-ci.org/display/JENKINS/CIFS-Publisher+Plugin
+clang-scanbuild-plugin = https://github.com/jenkinsci/clang-scanbuild-plugin/commit/a6d57b67f6fbd0a9893ecf6436c54ecb670d5829
+cloudbees-deployer-plugin = https://wiki.jenkins-ci.org/display/JENKINS/Cloudbees+Deployer+Plugin
+cloudbees-disk-usage-simple-plugin = https://github.com/jenkins-infra/update-center2/pull/179
+cloudbees-enterprise-plugins = https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Jenkins+Enterprise
+cloudbees-registration = https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Registration+Plugin
+codescan = https://github.com/jenkins-infra/update-center2/pull/299
+copyarchiver = https://wiki.jenkins-ci.org/display/JENKINS/CopyArchiver+Plugin
```
timja, line 25:
```suggestion
configuration-as-code-support = https://github.com/jenkinsci/configuration-as-code-plugin/releases/tag/configuration-as-code-1.18
copyarchiver = https://wiki.jenkins-ci.org/display/JENKINS/CopyArchiver+Plugin
```

daniel-beck:
> Configuration as Code support plugin is missing

FWIW I built this initial list from plugins that are not currently being distributed (i.e. "use case 1"), `configuration-as-code-support` is. The only reason to add it to the global list is to override the URL referenced ("use case 2").

jetersen:
use case 2 :)

daniel-beck:
I saw that in the suggested change and it makes sense.

I'm slightly more concerned about adding wrong information here than not including everything I could be, we can more easily iterate on the latter.

jetersen:
Sure thing, let's do it in a separate PR :)

daniel-beck:
Reduced effective length of the content added to JSON files in this PR from 6494 bytes to 4321 bytes (66%) by using git.io and shortened wiki URLs.

resources/deprecations.properties
daniel-beck: Added with a shortened URL.

daniel-beck:
> Sure thing, let's do it in a separate PR :)

To clarify, I'm asking for reviews primarily to confirm the content added is correct, not that it is complete (it is not, didn't add very unpopular plugins).

Added `configuration-as-code-support` in 61ba869cc6950b7e1054f1148db8f43dfafd2cc3

zbynek:
The Wiki extractor is not working properly for the shortened URLs, so the plugin site won't be able to show more than a link to the wiki (the short one), which expands to the long one and that in turn redirects to the plugin site (see https://plugins.jenkins.io/parameter-separator/ as an example). This could be fixed in the extractor, but is it worth the 2KiB?

daniel-beck:
@zbynek 

> the plugin site won't be able

This isn't plugin URL, it's deprecation URL. Why would that be inlined on the plugin site?

zbynek:
@daniel-beck my bad, sorry :facepalm:

daniel-beck:
I mean, I can see a possible future where we have placeholders for suspended plugins on the plugin site (it's a bit annoying it immediately goes to 404 today), but this can be re-evaluated then.

daniel-beck:
Ideally we'd get this merged early next week (Mon/Tue) so we'll get the core change merged towards the end of the week 😃

PR #346, opened 2020-02-26
Replace JS-Libs plugin URLs + mark some plugins deprecated

oleg-nenashev:
This change replaces plugin wiki links in https://github.com/jenkinsci/js-libs by redirects. All Wiki pages are stubs like below, so it should not be a problem. Changelogs will be lost, but they are empty stubs anyway

Some plugins were also marked as deprecated to document what README says. Some plugins have 200k+ installations though, so maybe we want to reconsider that.

image


oleg-nenashev:
CC @tfennelly

fqueiruga:
It makes sense to me to deprecate them, +1

oleg-nenashev:
> This will be fun. jquery-detached and momentjs are installed by the setup wizard by default.

I could remove the labels to retain the status quo

daniel-beck:
I mean, are they deprecated or not? If they are, the label should go in. If they're not, it should not.

oleg-nenashev:
> I mean, are they deprecated or not? If they are, the label should go in. If they're not, it should not.

I have no idea. They are kinda abandoned for sure, but they are de-facto used in other semi-abandoned functionality. I would rather prefer to retain the status quo than frighten users by deprecation notices.

CC @tfennelly @halkeye who could comment

oleg-nenashev:
Let's kick the fun part down the road

fqueiruga:
> This will be fun. jquery-detached and momentjs are installed by the setup wizard by default.

Not anymore, not since 2.217 :)

daniel-beck:
> If you mean used by the wizard, not anymore, not since 2.217!

Core cannot depend on _plugins_, which we're discussing here.

https://plugins.jenkins.io/pipeline-stage-view/ has dependencies on all these obsolete plugins.

Obligatory reference to https://github.com/jenkinsci/workflow-aggregator-plugin/pull/22 which still means everyone who installs Pipeline will get this stuff installed.

zbynek:
> 
> https://plugins.jenkins.io/pipeline-stage-view/ has dependencies on all these obsolete plugins.
> 

Even if you install pipeline plugins individually without the aggregator and pipeline view,  https://plugins.jenkins.io/workflow-cps/ still adds 2 of the JS libs. So thanks for reverting the deprecation. Wondering how hard would it be to replace these with @uhafner's new jQuery plugin.

oleg-nenashev:
I will merge it as is and leave the deprecation part to JSLibs and Pipeline plugin maintainers

PR #347, opened 2020-02-26
Permit the API Plugin label

oleg-nenashev:
The change adds official support for the https://github.com/topics/jenkins-api-plugin GitHub topic which represents the "API Plugin" term we were using recently for plugins which do not provide their own user-facing functionality. 

Problem: it collides with https://plugins.jenkins.io/ui/search/?labels=library which used to be used for such purpose in Wiki. IMHO "library" is a bit confusing, e.g. see https://github.com/topics/jenkins-library which refers Pipeline or Groovy script libraries. Needs discussion with interested parties to decide.

CC @uhafner @jglick @bitwiseman @jeffret-b @jvz @kuisathaverat @wadeck  who maintain API plugins from the lists

### DIFF
--- a/src/main/resources/allowed-labels.properties
+++ b/src/main/resources/allowed-labels.properties
@@ -9,6 +9,7 @@ administrative-monitor
 agent
 analysis
 android
+api-plugin
 aws
 bitbucket
 bitbucket-server


jglick:
I took the liberty of adding the label to a number of applicable plugins I have write access to. Some plugins which I know should have it but which I am not permitted to update:

* https://github.com/jenkinsci/kubernetes-credentials-plugin
* https://github.com/jenkinsci/branch-api-plugin
* https://github.com/jenkinsci/scm-api-plugin
* https://github.com/jenkinsci/credentials-plugin
* https://github.com/jenkinsci/script-security-plugin
* https://github.com/jenkinsci/ssh-credentials-plugin
* https://github.com/jenkinsci/structs-plugin
* https://github.com/jenkinsci/code-coverage-api-plugin
* https://github.com/jenkinsci/display-url-api-plugin
* https://github.com/jenkinsci/node-iterator-api-plugin
* https://github.com/jenkinsci/jenkins-plugin-runtime.rb

Some which I am not personally familiar with but am guessing should have it (a bunch of these are by @uhafner):

* https://github.com/jenkinsci/gitlab-api-plugin
* https://github.com/jenkinsci/azure-commons-plugin
* https://github.com/jenkinsci/forensics-api-plugin
* https://github.com/jenkinsci/analysis-model-api-plugin
* https://github.com/jenkinsci/mapdb-api-plugin
* https://github.com/jenkinsci/handy-uri-templates-2-api-plugin
* https://github.com/jenkinsci/master-to-master-api-plugin
* https://github.com/jenkinsci/plugin-util-api-plugin
* https://github.com/jenkinsci/mysql-api-plugin
* https://github.com/jenkinsci/data-tables-api-plugin
* https://github.com/jenkinsci/jnr-posix-api-plugin
* https://github.com/jenkinsci/jna-posix-api-plugin
* https://github.com/jenkinsci/postgresql-api-plugin
* https://github.com/jenkinsci/bootstrap4-api-plugin
* https://github.com/jenkinsci/jquery3-api-plugin
* https://github.com/jenkinsci/h2-api-plugin
* https://github.com/jenkinsci/popper-api-plugin
* https://github.com/jenkinsci/echarts-api-plugin

MarkEWaite:
I added the topic to the git-client-plugin as well.

uhafner:
I added the topic for all of these plugins that I own.

oleg-nenashev:
I updated all plugins listed by @jglick . We are likely miss some more plugins, but it is a good start.
One thing is that there is some topic deviation. I used https://github.com/topics/jenkins-api-plugin the while @uhafner used https://github.com/topics/api-plugin . It is not a problem for the Plugin site, but search for topics on GitHub will not return all results

uhafner:
Ah, I do not have a preference. The [file](https://github.com/jenkins-infra/update-center2/pull/347/files) in this PR indicates that the label is `api-plugin` so I used that one. We should all use the same if possible. So I can change to jenkins-api-plugin...

MarkEWaite:
I used api-plugin as well because that is the label in the approved label list.  I removed the labels that were not on the approved list from the three plugins that I touched.  Nice cleanup to be reminded that there is an approved list that we should prefer to use first.

PR #348, opened 2020-02-26
Whitelist some technology labels

oleg-nenashev:
This pull request whitelists plugin labels related to technology topics, e.g. Git, Docker, Kubernetes, Windows, Google Cloud, Azure, etc. For each of the suggested labels there are at least 5 plugins which match them (I still need to update GitHub topics), so I believe this categorization will be useful to users. Later we can also add some filtering views based on that.

Jenkins Packaging service project idea could benefit from it as well

CC @MarkEWaite @LinuxSuRen


src/main/resources/allowed-labels.properties
```diff
@@ -21,21 +22,27 @@ cluster
...
+configuration-as-code
 database
 deployment
 devops
 dotnet
+docker
 emailext
 external
 file
+git
```
MarkEWaite: Could we also include `gitea` as a whitelisted topic so that all branch source providers have a topic?
```suggestion
git
gitea
```

oleg-nenashev: @MarkEWaite "For each of the suggested labels there are at least 5 plugins which match them" - it was my criteria for the list. Gitea does not match it right now. We could add it later if there is a need in such label. We still have a search working well

MarkEWaite: I'm fine with this pull request without gitea.

I would like to better understand the criteria that decide if a plugin receives a specific label?

For example, would the following plugins receive the GitHub label?

* [Basic branch build strategies](https://plugins.jenkins.io/basic-branch-build-strategies/)
* [Pipeline multibranch](https://plugins.jenkins.io/workflow-multibranch/)
* [Skip notifications trait](https://plugins.jenkins.io/skip-notifications-trait/)
* [Generic Webhook Trigger](https://plugins.jenkins.io/generic-webhook-trigger/)
* Git plugin
* Git client plugin

If those are labeled for GitHub, then I think they should also be labeled for Gitea.  If they are not labeled for GitHub, then I'm not clear how the GitHub label will be any different than searching the plugins site for the keyword 'GitHub'.

Are you envisioning that the label will applied to plugins that are strongly recommended for use in the target environment (describing a "minimum set of plugins") or that it would be applied to plugins that might be optionally considered for use in the target environment (guiding the user as they consider their choice to install plugins)?

If strongly recommended, then Gitea may not have 5 plugins that match the criteria.  If optionally applied, then I believe there are more than 5 plugins that are relevant to a Jenkins server using Gitea.

oleg-nenashev: I think it is a subject for a separate discussion and a doc outside this PR. Jenkins Wiki labeling was a Wild West, and I would rather like to recover the status quo. Documenting best practices is important, but it is rather a follow-up

oleg-nenashev:
@MarkEWaite would you be interesting to start a labeling policy proposal in the Docs SIG or development mailing list? Your comments are already a good starting point

oleg-nenashev:
I haves started a thread about labeling: https://groups.google.com/forum/#!topic/jenkinsci-dev/DRzc6TatYSc

MarkEWaite:
Happy to start a separate thread or to include the topic in the thread you've already created.  Do you have a preference which of those it should be?

PR #349, opened 2020-02-28
Add localization as a label

LinuxSuRen:


daniel-beck:
It would probably help demonstrate usefulness by adding labels to appropriate plugins into the properties file: `translation`, `localization-support`, `localization-zh-cn`, are there others?

LinuxSuRen:
You mean put those into `src/main/resources/label-definitions.properties`?

PR #350, opened 2020-03-01
Add atlassian plugin labels

warden:


daniel-beck:
Please provide examples for these labels. @MarkEWaite proposed having five plugins at least for each label in https://groups.google.com/d/msg/jenkinsci-dev/mT4aVq4cJD0/aeCtebXgHAAJ

`atlassian` in particular is difficult, as it could imply an authorship/ownership where there is none. I also don't see the benefit to users there, who are probably more interested in specific `$productName` than `atlassian`.

warden:
@daniel-beck got it and removed `Atlassian`, I think the other labels are known enough.

src/main/resources/allowed-labels.properties
```diff
@@ -12,6 +12,7 @@ android
 api-plugin
 aws
 azure
+bamboo
```
MarkEWaite: I only found one Jenkins plugin that seems to match this tag:
* [Bamboo pipeline step](https://plugins.jenkins.io/pipeline-bamboo/)

Since Bamboo is a continuous integration server, I would not expect to have many Jenkins plugins related to Bamboo.  

I think it is premature to whitelist a `bamboo` label until we know there are multiple plugins that would use the label

src/main/resources/allowed-labels.properties
```diff
@@ -22,6 +23,7 @@ cloud
 cluster
 cmp
 codebuild
+confluence
```
MarkEWaite: I found three plugins that seem like good candidates for the confluence label:

* [Confluence publisher](https://plugins.jenkins.io/confluence-publisher/)
* [Release helper](https://plugins.jenkins.io/release-helper/)
* [Docktor](https://plugins.jenkins.io/doktor/) 

There is also a reference to a commercial product from "Cloudpipes" but no reference to Cloudpipes on any Jenkins site that I can see.

Are there other Confluence related plugins that I have missed?

src/main/resources/allowed-labels.properties
```diff
@@ -38,6 +40,7 @@ gitlab
 groovy-related
 google-cloud
 ios
+jira
```
MarkEWaite: Plugins that seemed like good candidates for this label include:

* [Atlassian Jira software cloud](https://plugins.jenkins.io/atlassian-jira-software-cloud/)
* [Jira](https://plugins.jenkins.io/jira/)
* [Test management for Jira](https://plugins.jenkins.io/tm4j-automation/)
* [X-ray connector for Jira](https://plugins.jenkins.io/xray-connector/)
* [Jira pipeline steps](https://plugins.jenkins.io/jira-steps/)
* [Jira issue updater](https://plugins.jenkins.io/jenkins-jira-issue-updater/)
* [Zephyr for Jira test management](https://plugins.jenkins.io/zephyr-for-jira-test-management/)
* [QMetry for Jira test management](https://plugins.jenkins.io/qmetry-for-jira-test-management/)
* [Bitbucket Jira validator SCM filter](https://plugins.jenkins.io/bitbucket-scm-filter-jira-validator/)
* [GitHub Jira Validator SCM filter](https://plugins.jenkins.io/github-scm-filter-jira-validator/)
* [Jira integration for Blue Ocean](https://plugins.jenkins.io/blueocean-jira/)
* [Jira test result reporter](https://plugins.jenkins.io/JiraTestResultReporter/)
* [iTMS for Jira](https://plugins.jenkins.io/itms-for-jira/)
* [Release Helper](https://plugins.jenkins.io/release-helper/)
* [Generic Webhook trigger](https://plugins.jenkins.io/generic-webhook-trigger/)

Jira is a great candidate for a label.  Are there plugins other than those which should receive the `jira` label?

src/main/resources/allowed-labels.properties
oleg-nenashev: https://github.com/jenkinsci/confluence-pipeline-steps-plugin but it is archived

PR #351, opened 2020-03-02
Whitelist the email plugin label

oleg-nenashev:
https://github.com/search?q=topic%3Aemail+org%3Ajenkinsci&type=Repositories . There are actualy more plugins, but looks like we hit an issue with forked repositories which do not get listed. FYI @daniel-beck

### DIFF
--- a/src/main/resources/allowed-labels.properties
+++ b/src/main/resources/allowed-labels.properties
@@ -29,6 +29,7 @@ deployment
 devops
 dotnet
 docker
+email
 emailext
 external
 file


daniel-beck:
I would expect only https://plugins.jenkins.io/email-ext/ and https://plugins.jenkins.io/emailext-template/ to be eligible for `emailext`. Can probably be removed. CC @slide

slide:
Yes, I would say removing emailext would be good.

PR #353, opened 2020-03-10
[INFRA-2513] - Update allowed-labels.properties

jtnord:
do not allow a slave label

someone with access to all infra could find/replace this label with agent if they where so inclined.


oleg-nenashev:
If we take `agents` or something similar as a new label, again I will do the GitHub legwork. Created https://issues.jenkins-ci.org/browse/INFRA-2513 as a placeholder

jtnord:
changed all to `agent` in the properties file.  looks a little strange as the prior label was pluralised (slaves) and now it is not (agent)  but other things are singular (report vs reports / parameter vs parameters)

daniel-beck:
@jtnord Are you still interested in this change?

jtnord:
> @jtnord Are you still interested in this change?

I do not know enough about the UC to know what needs changing and where to move this forward.

This was an attempt to keep aligning things better given the project decided to move to agents.

oleg-nenashev:
If you are fine @daniel-beck @jtnord , I would like to keep it open. I want to deliver an engine for label aliases, and I will do it on the top of this PR

daniel-beck:
Works for me.

oleg-nenashev:
I will be unable to work on a new engine anytime soon. It would be needed for removing "slave" terms from labels and Artifact IDs. Maybe @MarkEWaite or @markyjackson-taulia would be interested to facilitate this change in the community

markyjackson-taulia:
I can work on this. I will sync with you @oleg-nenashev

daniel-beck:
> and Artifact IDs

That's quite some scope creep here.

daniel-beck:
Given the lack of followup in several years I guess this can be closed after all.

We can always restart this if we want, then with a corresponding core change.

PR #355, opened 2020-03-13
Add issue tracker to update center

halkeye:
Superseeds #331 

Pulled in master from #331 
Switched to using graphql for pulling issuetracker conditional
Switched to using graphql to get all the repo names

The new class I created might need a rename but i'm happy with this output.

```
 ☎  ~  …  update-center2  add-issue-tracker-to-update-center  2+  %    cat output/stable/update-center.actual.json | jq '.plugins["blueocean-web"]["issueTracker"]'
{
  "name": "Jenkins Jira",
  "url": "https://issues.jenkins-ci.org/"
}
 ☎  ~  …  update-center2  add-issue-tracker-to-update-center  2+  %    cat output/stable/update-center.actual.json | jq '.plugins["configuration-as-code"]["issueTracker"]'
{
  "name": "GitHub Issues",
  "url": "https://github.com/jenkinsci/configuration-as-code-plugin/issues"
}
```


halkeye:
btw, i was super lazy and just updated the test graphql files to always return false instead of generating them again

```
const fs = require('fs');
const dir = 'src/test/resources/';
["github_graphql_null.txt", "github_graphql_Y3Vyc29yOnYyOpHOA0oRaA==.txt"].forEach(function(file) {
  const data = JSON.parse(fs.readFileSync(`${dir}${file}`).toString());
  data.data.organization.repositories.edges.forEach(({ node }) => {
    node.hasIssuesEnabled = false;
  });
  fs.writeFileSync(`${dir}${file}`, JSON.stringify(data, null, 2))
})
```

daniel-beck:
Conflicted after https://github.com/jenkins-infra/update-center2/pull/343.

Also looking over this superficially, it seems to drop the support of unauthenticated GitHub use that I just fully restored in #343. If so, please restore that behavior when no credentials are set.

halkeye:
ah, now there's a conflict. I think 343 wasn't merged, I had assumed it was already. I'll look into it.

daniel-beck:
Sorry about that, given the importance of this tool for the Jenkins project and the shittiness of the test coverage every deployment is quite the operation to ensure nothing unexpected happens. So stuff is delayed a lot.

halkeye:
Also updated dom4j in an effort to fix the too many files open issue in linux

Turns out it didn't actually solve the problem, i still have it about 50% of the time.

As from irc:
```
19:10 <halkeye> also danielbeck have you ever looked into a too many open files error in update center?
20:05 <halkeye> looks like MavenArtifactSource is leaking a handle
20:05 <halkeye> i think specifically JarFile jar = new JarFile(f);
20:23 <halkeye> oh updating java i can run it again, something is weird, i'm not 100% certain its consistent
```

daniel-beck:
What about plugins that define `issueTracking` in `pom.xml`? Looks like we're just ignoring that here.

https://github.com/search?l=Maven+POM&q=org%3Ajenkinsci+issuemanagement&type=Code

halkeye:
> What about plugins that define `issueTracking` in `pom.xml`? Looks like we're just ignoring that here.

Yea I brought that @timja when he first wrote it. Its not hard to fix, I just don't know of any to test with offhand.

timja:
Didn’t realise that any plugins were defining it, I wrote the code to get it from the Pom but wasn’t sure if it was a great idea as it would be hard to get plugins to include it and very few do

daniel-beck:
Ideally (IMO), we use pom.xml first, and if undefined, fall back to the current heuristic.

timja:
> Ideally (IMO), we use pom.xml first, and if undefined, fall back to the current heuristic. WDYT?
> 
> The only problem seems to be that a change of issue tracker would require a new release, but that is a pretty big operation anyway.

sounds fine

daniel-beck:
Please note that merge of the overhaul PR is likely coming this week, so probably useful to hold off work on improving this PR until then.

daniel-beck:
FYI I merged #265.

daniel-beck:
Ideally this class would be updated to use gson instead of json-lib so we can get rid of that dependency, as we're no longer using it to generate output.

halkeye:
I can do that when I update it. It's just low priority for me cause last I heard GitHub was still not officially supported for plugins.

halkeye:
But when it is, and/or when plugin site updates go live it'll bump up

daniel-beck:
Superseded by #461 (which doesn't migrate to Gson 😢 )

PR #356, opened 2020-03-14
Add popularity indicator

daniel-beck:
Upstream of https://github.com/jenkinsci/jenkins/pull/4588

I am currently looking at improving how plugins are presented in the plugin manager in Jenkins.

One problem is the lack of a useful sort order in the "Available" tab. Even filtering for a relevant keyword like "Git"  will have tons of results, the most relevant plugins (like "Git Plugin") hidden way down the list.

Currently the only available metric in Jenkins for a default sort order would be release date, which will move unmaintained plugins to the bottom, but often, the most recently released plugins aren't the most relevant.

So add a general "popularity" field for plugins that can be used (directly or indirectly) to sort available plugins in Jenkins. I'm leaving the definition explicitly vague to allow later changes that take into account more metrics, like giving a boost to "Trending" plugins (i.e. when much more popular now than a while ago, see right column on https://plugins.jenkins.io/). This also means that the current range of 0...1 will not always be true.

This seems to add an average of 34 bytes to the output, or 58kb for ~1700 plugins. That's about a 3% size increase for the entire JSON file.


halkeye:
now who is conflicting :)

This might actually be the last thing that actually kept the plugin api server around, the readme parsing could easily be moved to the static site generator, then less moving parts.

I like it, the in app search is pretty frustrating.

daniel-beck:
> This might actually be the last thing that actually kept the plugin api server around

Not really, as this is way less powerful – with the current implementation, you'd have to give up the "Trending" category.

fqueiruga:
TBH I'm not sure what's better UX-wise. While I've definitely felt the pain of searching for the Git plugin before, I'm not sure that doing without alphabetical order would be better.

Now that you've done great work setting a popularity value, maybe we could add some sort of filtering by that field, instead of ordering. Maybe have 3 options like "Popular, commonly used, barely used" (I'm sure there are better categories). WDYT?

Other than that, I'm neutral to these changes, I don't have an informed opinion on this

oleg-nenashev:
> This might actually be the last thing that actually kept the plugin api server around, the readme parsing could easily be moved to the static site generator, then less moving parts.

I would keep it for now. We might later need it for advanced queries (e.g. GraphQL support). We can deduplicate on the frontend side if needed

PR #361, opened 2020-04-02
Suspend kubesphere-extension releases with incremental dependencies

daniel-beck:


daniel-beck:
FTR the plugin cannot be installed on a real instance, since the dependency is never satisfied, and Jenkins refuses to load it.

daniel-beck:
CC @runzexia

PR #368, opened 2020-04-20
Last affected version pattern changed to 1.9.1

pv-smartbear:
Fix for given security issue was given in v1.10 but the message is showing for v2.0 also. Changed pattern to remove version v2.x out of this issue.


pv-smartbear:
@daniel-beck Any update on this?

daniel-beck:
Due to a bug in 2.233 we're running UC generation off an older commit to compensate. That'll be reverted probably today now that 2.234 is out, and we'll resume merging blacklist changes.

daniel-beck:
Sorry about this BTW.

We autogenerate these regexes based on affected/fixed metadata, and only recently found out that if there have been further releases since the fix, that those are then considered affected, rather than fixes.

pv-smartbear:
No problem.

PR #370, opened 2020-04-22
Blacklist azure-credentials 3.0.1

timja:
https://issues.jenkins-ci.org/browse/JENKINS-62011 causes wide spread breakage in the azure plugins

cc @xuzhang3 for approval

### DIFF
--- a/src/main/resources/artifact-ignores.properties
+++ b/src/main/resources/artifact-ignores.properties
@@ -388,3 +388,6 @@ configuration-as-code-1.37
 # https://github.com/jenkinsci/kubesphere-extension-plugin/pull/4#pullrequestreview-386832816
 kubesphere-extension-0.1.0
 kubesphere-extension-0.2.0
+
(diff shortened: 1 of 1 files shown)


daniel-beck:
FTR we're not currently building update sites from latest master due to https://github.com/jenkinsci/jenkins/pull/4675; so merging this will have no effect without a corresponding configuration change to trusted.ci.

timja:
> FTR we're not currently building update sites from latest master due to [jenkinsci/jenkins#4675](https://github.com/jenkinsci/jenkins/pull/4675) (not sorting by popularity is better than reverse sort); so merging this will have no effect without a corresponding configuration change to trusted.ci.

@daniel-beck is it running off a specific branch or is it a commit?

timja:
4.0.0 has been released, I'll follow up with xuzhang about the breakage, but at least users aren't getting this picked up in a patch version

PR #372, opened 2020-04-27
Suspend plugin from listing in Update center

nikita-bedmutha:
GA of Qualys API Security Module is unfortunately delayed and hence need to suspend its plugin until we have the module available publicly.

The github link of this plugin - https://github.com/jenkinsci/qualys-api-security-plugin


daniel-beck:
FTR requester is a committer to the plugin.

daniel-beck:
@nikita-bedmutha To clarify, you want to stop publication of the entire plugin, whose only release so far was in October?

nikita-bedmutha:
Yes @daniel-beck, we had only one version release done so far, and we want to hide that entirely until our product module release.

PR #377, opened 2020-05-18
Removing Fortify Cloudscan plugin from distribution

stevespringett:
Removing Fortify Cloudscan plugin from distribution in favor of vendor supported plugin


daniel-beck:
On-holding this until after https://github.com/jenkins-infra/update-center2/pull/344 is merged so we can define a deprecation for this.

Also needs a rebase onto current master after the overhaul.

daniel-beck:
Superseded by #392.

PR #382, opened 2020-05-25
[JENKINS-62332] Add information about latest plugin version

daniel-beck:


src/main/java/io/jenkins/update_center/LatestPluginVersions.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public static void initialize(@Nonnull MavenRepository repository) throws IOException {
+        instance = new LatestPluginVersions(repository);
+    }
+
+    public static void initializeEmpty() {
+        instance = new LatestPluginVersions(Collections.emptyMap());
+    }
+
+    @CheckForNull
```
varyvol, line 65: Why checkForNull if you are later requiring it to be non null?

daniel-beck: Good point, the original implementation of disabling this feature didn't set an instance, rather than call `initializeEmpty()`. This is a leftover from that.

PR #384, opened 2020-06-03
Stop distributing play-autotest plugin

daniel-beck:
Apparently the plugin doesn't work anymore anyway.

https://www.jenkins.io/security/advisory/2020-06-03/


Blakehale999 (commented):
Nerge

PR #385, opened 2020-06-05
blacklist configuration-as-code-support

jetersen:
Since the release of 1.18 which is a year ago: https://github.com/jenkinsci/configuration-as-code-plugin/releases/tag/configuration-as-code-1.18
The code that was moved to other plugins where it belonged have had time to adopt.

I suggest we blacklist.

cc @timja

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -28,6 +28,7 @@ cloudbees-registration           # "This feature is no longer relevant." -- http
 cloudbees-disk-usage-simple-plugin # renamed to cloudbees-disk-usage-simple several years ago
 ColumnPack-plugin                # renamed to ColumnsPlugin
 ConfigurationSlicing             # renamed into configurationslicing, and this double causes a check out problem on Windows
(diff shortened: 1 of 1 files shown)


jetersen:
We should properly override the plugins page as well 😓
https://plugins.jenkins.io/configuration-as-code-support/

daniel-beck:
FTR deprecations entry here: https://github.com/jenkins-infra/update-center2/blob/aac997727c68b9a509492e595a3157d3e286f857/resources/deprecations.properties#L41

jetersen:
Not sure if the plugin site could parse that url 🤔
We would properly need a separate readme somewhere...

daniel-beck:
> plugin site

Suspended plugins have no plugin site page. Never had.

oleg-nenashev:
> Suspended plugins have no plugin site page. Never had.

Indeed. We could add tombstone pages for removed/depublished plugins in the future, but it is not what we are doing now

jetersen:
I am fine if the plugin disappear from plugins site, once blacklisted.

PR #387, opened 2020-06-06
Update dependencies

daniel-beck:


pom.xml
```diff
@@ -214,11 +209,29 @@
     <dependency>
       <groupId>com.google.code.findbugs</groupId>
       <artifactId>annotations</artifactId>
-      <version>3.0.0</version>
+      <version>3.0.1u2</version>
```
MarkEWaite, line 212: https://mvnrepository.com/artifact/com.google.code.findbugs/annotations suggests that 3.0.1 is newer than [3.0.1u2](https://mvnrepository.com/artifact/com.google.code.findbugs/annotations/3.0.1u2).  However, that seems "backwards" to me, since I would expect a ".1u1" and a ".1u2" to be newer than a ".1".  All three were released Oct 9, 2015.  No objections from me in either case, just surprising to see a version change on a mature library like this one.

daniel-beck: Good point, haven't even considered that.

https://repo1.maven.org/maven2/com/google/code/findbugs/annotations/ indicates u2 is the newest.

The difference between 3.0.1 and 3.0.1u1 is basically https://github.com/findbugsproject/findbugs/commit/2f9d672b1f4469074ed23b728fb2ec1f55766dd8#diff-d2b359d7422403111420fc606058a061 (inclusion of `edu/umd/cs/findbugs/annotations/*Warning.java`)

The difference between 3.0.1u1 and 3.0.1u2 is https://github.com/findbugsproject/findbugs/commit/725be6e2c911f6bc7c7b7589911b9db04625972f#diff-78590aa3b479ed0ff7b772a18df8bbab (removal of `Priorities`, minor changes in enums).

zbynek: This has no impact on runtime, so either option seems fine. It could be replaced with
https://mvnrepository.com/artifact/com.github.spotbugs/spotbugs-annotations 4.0.4 in another PR.

oleg-nenashev: Yes, SpotBugs is a way to go

PR #388, opened 2020-06-06
Use gson to parse popularity metadata from stats.jenkins.io

daniel-beck:
The less json-lib, the better 😁


daniel-beck:
Superseded by #394.

PR #389, opened 2020-06-16
Allow the "theme" plugin label in the Update Center

oleg-nenashev:
https://github.com/search?q=topic%3Atheme+org%3Ajenkinsci+fork%3Atrue

CC @timja @tobix

### DIFF
--- a/resources/allowed-github-topics.properties
+++ b/resources/allowed-github-topics.properties
@@ -80,6 +80,7 @@ spotinst
 stash
 test
 textfile
+theme
 trigger
 ui
 upload


oleg-nenashev:
> Ideally would have a companion change to core localizations:

I think we should move it to the update center TBH.
Subject for future discussion.

daniel-beck:
> I think we should move it to the update center TBH.

I disagree. Bloats the JSON and/or will not support localizations. Having unlocalizable security warnings is already bad enough.

PR #390, opened 2020-06-21
Mark plugins deprecated by maintainers as such

zbynek:
These plugins all mention that they are deprected in their documentation.

If the preference is to use labels on repos, somebody with access can do that for these repos instead:
* https://github.com/jenkinsci/fortify-cloudscan-plugin
* https://github.com/jenkinsci/multiple-scms-plugin
* https://github.com/jenkinsci/scm-sync-configuration-plugin
* https://github.com/jenkinsci/coverity-plugin
* https://github.com/jenkinsci/rrod-plugin
* https://github.com/jenkinsci/eggplant-plugin
* https://github.com/jenkinsci/persistent-build-queue-plugin
* https://github.com/jenkinsci/matrixtieparent-plugin

For two of the plugins that seems impossible
* https://github.com/jenkinsci/vsts-cd-plugin-plugin is not a valid repo
* https://github.com/jenkinsci/appscan-plugin is re-used for the replacement plugin


daniel-beck:
https://github.com/jenkinsci/jenkins/pull/4073 😭

daniel-beck:
I labeled all repos in the first list except for the currently archived fortify-cloudscan-plugin as `deprecated`.

zbynek:
@daniel-beck thanks for setting the labels! And I've seen your anniversary post in that other PR. For the three plugins where label can't be set on a repo (`ibm-application-security`, `fortify-cloudscan`, `vsts-cd`) -- should I add it as a label or use the new deprecation format?

daniel-beck:
@zbynek Depends on whether you want the deprecation link to point to a different URL than the "wiki" URL. See https://github.com/jenkins-infra/update-center2#deprecations

zbynek:
@daniel-beck thanks. Since there's no URL describing the deprecation better than the docs (that I know of), I just kept the label for those 3 plugins.

PR #396, opened 2020-07-06
[SECURITY-1014] Match vulnerable Libvirt Agents

bgermann:
The SECURITY-1014 vulnerabilities were fixed in Libvirt Agents plugin
version 1.8.6.  Match only the vulnerable versions in order to warn.


resources/warnings.json
```diff
@@ -6377,7 +6377,7 @@
     "versions": [
       {
         "lastVersion": "1.8.5",
-        "pattern": ".*"
+        "pattern": "1\\.[0-7].*|1\\.8.[0-5]"
```
daniel-beck, line 6380: Better never release version 1.10.

I recommend you model your pattern after the existing ones. My recommendation:

    (1[.]7|1[.]8[.][1345])(|[.-].+)|(1[.]8)

bgermann: Okay, I updated the pattern according to your suggestion.

daniel-beck:
@bgermann The CSRF vulnerability that is part of SECURITY-1014 (2) has not been resolved.

Please see the documentation at https://www.jenkins.io/doc/developer/security/form-validation/

bgermann:
Okay, let me fix that. I'll open a new PR once it is fixed.

PR #398, opened 2020-07-14
Add label for Overall Manage Permission Enabler plugin

escoem:
https://github.com/jenkinsci/manage-permission-plugin

### DIFF
--- a/resources/label-definitions.properties
+++ b/resources/label-definitions.properties
@@ -592,6 +592,7 @@ macstadium-orka=cluster slaves ios
 mail-watcher-plugin=notifier
 mailcommander=misc trigger
 maintenance-jobs-scheduler=misc
+manage-permission=user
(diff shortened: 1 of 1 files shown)


daniel-beck:
Why not https://github.com/jenkins-infra/update-center2#github ?

escoem:
> Why not https://github.com/jenkins-infra/update-center2#github ?

I'm also setting it as topic but I have seen other plugins like https://github.com/jenkinsci/extended-read-permission-plugin with topics (`jenkins`, `jenkins-plugin`, `permission`, `advance`) different than what they have athttps://github.com/jenkins-infra/update-center2/blob/master/resources/label-definitions.properties#L331 (`user`), that's why I'm also sending this PR.

daniel-beck:
These plugins pre-date support for GitHub labels.

PR #399, opened 2020-07-16
Suppress accidental BFA release

rsandell:


daniel-beck:
Public for one hour, I wonder what the install stats will say next month.

daniel-beck:
5 installs according to stats of people now stuck on 2.27.

PR #400, opened 2020-07-21
WIP: Ignore alauda's plugins due to closed publish flow

LinuxSuRen:


daniel-beck:
@LinuxSuRen Could you provide more context for this?

LinuxSuRen:
@daniel-beck Thanks for your quick response. This PR still under working in progress. I'll let you know once it's ready.

LinuxSuRen:
My team made a decision about our plugins. Currently, our plugins can only resolve our problems. So I guess there are no other users who are using the base on the download count number. Furthermore, we already have a private plugins publish flow base on [Custom war Packager](https://github.com/jenkinsci/custom-war-packager). As you can see, we don't release them to update.jenkins.io for several months. Thanks to @oleg-nenashev taking the initiative to start CWP project.

In order to let this PR be more clear. I address my team's core members below:
@cizezsy @chengjingtao

One more question, @daniel-beck do I need to remove the security warnings from [warnings.json](https://github.com/jenkins-infra/update-center2/blob/d06900a10996bc2f366edc8cbb808d62d2ec6d9d/resources/warnings.json#L6790) in this PR?

daniel-beck:
> do I need to remove the security warnings from warnings.json in this PR?

No, these plugins have previously been published by the Jenkins project, so these warnings will remain.

daniel-beck:
Given the nature of this change, a deprecation should probably be added, too: https://github.com/jenkins-infra/update-center2#deprecations

In case anyone else does use these plugins.

LinuxSuRen:
> > do I need to remove the security warnings from warnings.json in this PR?
> 
> No, these plugins have previously been published by the Jenkins project, so these warnings will remain.

You made a good point.

To keep a historical record for those plugins, an alternative solution is adding a deprecation label and archive it. But I'm totally okay with removing them from `jenkinsci` GitHub org.

oleg-nenashev:
@LinuxSuRen Hi, any updates here?

LinuxSuRen:
@daniel-beck Thanks for the clarification.

PR #405, opened 2020-07-22

andreashoyerh2m:
Depublish "sonargraph-plugin". It worked with Sonargraph version 7 which is now discontinued. Sonargraph version 8/9/10 is supported by "sonargraph-integration", which is the successor of "sonargraph-plugin".


oleg-nenashev:
Maybe it is better to mark the plugin as deprecated and archive it, but no strong opinion

andreashoyerh2m:
It should have been deprecated two years ago, whilst Sonargraph 7 was still 'working' theoretically, but we missed that. Now this plugin cant be used in anymore, and I think depublishing is the right thing to do.

daniel-beck:
@andreashoyerh2m Given ~2400 installations, it probably makes sense to also mark it as deprecated via resource file: https://github.com/jenkins-infra/update-center2#deprecations

Doing that will show a message to admins who have it installed.

andreashoyerh2m:
Marked it as deprecated

daniel-beck:
@andreashoyerh2m Looks like I wasn't clear: for a suspended plugin, the second option needs to be used. There's nowhere to attach labels to in the metadata.

daniel-beck:
@oleg-nenashev Per my comment, this wasn't done. The label will never be applied.

phuchau1989 (approved):

PR #411, opened 2020-08-01
Suspend distribution of sideex-1.0.0

daniel-beck:
Looks like both 1.0 and 1.0.0 were released and that creates ambiguity in what we publish:
https://github.com/jenkinsci/sideex-plugin/commits/master

From build logs and https://updates.jenkins.io/download/plugins/sideex/ it looks like 1.0 is currently being published at least fairly consistently, so choose that over 1.0.0.

FYI @kiam123, please don't create releases whose precedence is ambiguous (`X` and `X.0`).

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -411,6 +411,9 @@ elastest-1.0.0-beta1
 # ambiguous version numbers: 0.1.0-alpha-1 is preferred
 git-forensics-0.1-alpha-1
 
+# ambiguous version numbers: Prefer 1.0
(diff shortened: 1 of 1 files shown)


kiam123:
Hi, thanks for pointing out the issue, we wanted to keep the version 1.0.0, but were having trouble removing the previous version 1.0, we are still attempting to fix this issue

daniel-beck:
@kiam123 Just release 1.0.1. That'll take precedence.

kiam123:
@daniel-beck ok, we are preparing release 1.0.1

PR #416, opened 2020-08-03
Nest dynamic tier directories into common top-level directory

daniel-beck:
https://updates.jenkins.io/ is a mess 😮 So just hide most of it in one `dynamic/` top level directory.


daniel-beck:
Weird, the change seems to have caused quite the increase in execution time, from consistently just under 2 minutes so wildly fluctuating between that and more than 3 minutes.

Screenshot 2020-08-04 at 09 12 19

PR #421, opened 2020-08-07
Terminology update (blacklisted -> suspended)

zbynek:


daniel-beck:
Some occurrences were `blacklist` (active, present tense) and were changed to `suspended`, deliberate?

zbynek:
I changed it back to the present tense (but I think the comments make sense either way).

PR #422, opened 2020-08-10
Add sonargraph-plugin to the deprecation list

oleg-nenashev:
Follow-up to https://github.com/jenkins-infra/update-center2/pull/405 . CC @daniel-beck @andreashoyerh2m

### DIFF
--- a/resources/deprecations.properties
+++ b/resources/deprecations.properties
@@ -122,6 +122,8 @@ secret = https://wiki.jenkins.io/x/9ghSAg
 setenv = https://wiki.jenkins.io/x/YAtSAg
 # https://wiki.jenkins.io/pages/viewpage.action?pageId=63930505
 sonatype-ci = https://wiki.jenkins.io/x/iYDPAw
+# https://wiki.jenkins-ci.org/display/JENKINS/Sonargraph+Plugin
+sonargraph-plugin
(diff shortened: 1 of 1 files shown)


daniel-beck:
Needs a URL.

oleg-nenashev:
URL to what?

daniel-beck:
@oleg-nenashev An URL that ideally describes what's going on with the plugin (e.g. why deprecated, what users should use instead, etc.). Please see https://github.com/jenkins-infra/update-center2#deprecations

daniel-beck:
Note that all other comments in this file explain what the minified links point to. If you want to link to the wiki page (after disabling the redirect to the plugin site), the current comment makes sense, otherwise it does not. See https://github.com/jenkins-infra/update-center2/blob/e7265c093ebf681db938f73ca3c405df57d8084d/resources/deprecations.properties#L5

daniel-beck:
@oleg-nenashev Ping?

oleg-nenashev:
Pong. Feel free to amend and finalize it. You know my current situation. I
had a shitty week, and I have very difficult months ahead. I will announce
it later, but I will have to take a break in the most of community roles.

daniel-beck:
Closing in favor of #426. I have no idea what happens with an entry without a URL, and I don't intend to find out.

PR #434, opened 2020-08-22
Add fingerprint to allowed-github-topics

stellargo:
Add `fingerprint` to allowed-github-topics.properties

CC @oleg-nenashev @afalko @mikecirioli

### DIFF
--- a/resources/allowed-github-topics.properties
+++ b/resources/allowed-github-topics.properties
@@ -35,6 +35,7 @@ email
 emailext
 external
 file
+fingerprint
 git
 github
 gitlab


yJunS (approved):

zbynek:
For the record this label is only used by [one released](https://github.com/jenkinsci/redis-fingerprint-storage-plugin/)  and [one unreleased](https://github.com/jenkinsci/postgresql-fingerprint-storage-plugin) plugin. Not sure if it would make sense to add that label to the [Fingerprint plugin](https://plugins.jenkins.io/create-fingerprint/) (seems to be only relevant for freestyle jobs).

PR #440, opened 2020-09-09
Match 1.x+ rather than 1

MarkEWaite:
Jenkins 2.249.1 was incorrectly flagged as having a security issue


MarkEWaite:
Proposed change is incorrect and incomplete

PR #442, opened 2020-09-09
Skip sanity check for skipped generated versions

daniel-beck:
Amends #418.


MarkEWaite (approved):
I do not understand that code, but I assume this is resolving an update center issue and will help us.

PR #443, opened 2020-09-09
Include next higher LTS as a baseline for weekly dependencies

daniel-beck:
Right now, https://updates.jenkins.io/tiers.json indicates there are plugins depending on 2.236, 2.238, 2.240, 2.243, and 2.248, but we don't generate a 2.249.1 tier for LTS, even though this version has been released earlier today and is "visible" to this tool. `https://updates.jenkins.io/update-center.actual.json?version=2.249.1` redirects to 2.235.5.

This change should fix that by adding the lowest LTS release necessary when a plugin has a dependency on a weekly release.

I would guess this just wasn't noticeable recently because developers tend to depend on a ".1" release fairly quickly, masking this problem for all lines that were supported already when I introduced this change a few months ago.

### DIFF
--- a/src/main/java/io/jenkins/update_center/json/TieredUpdateSitesGenerator.java
+++ b/src/main/java/io/jenkins/update_center/json/TieredUpdateSitesGenerator.java
@@ -37,6 +37,7 @@ import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):

MarkEWaite:
I assume this will fix the case where my Jenkins 2.249.1 was unwilling to install Pipeline: Steps and Processes 2.36 because it requires 2.248.

PR #446, opened 2020-09-15
Add target=_blank for all links in plugin descriptions

daniel-beck:
Quite a few plugin descriptions contain an HTML link to the service's website or similar.

This change modifies these links to always have `target="_blank"`, and set `rel="noopener noreferrer"` in addition to `rel="nofollow"` (the default, doesn't really matter).


Wadeck (approved):
Not manually tested but the code seems fine 🐝

PR #450, opened 2020-09-25
Update particular version in artifact-ignores.properties

nikita-bedmutha:


daniel-beck:
Please provide an explanation for this change. It contradicts #372.

nikita-bedmutha:
> Please provide an explanation for this change. It contradicts #372.

@daniel-beck , thanks for reaching back, as quoted in #372, we had delay in release of our product module , which we are planning next week. So we want to enable back the plugin. However, during our automation for mvn release of jenkins plugins we had ended up mistakenly adding version 2.1 to this repo, hence need to hide that particular version.

daniel-beck:
@nikita-bedmutha What's wrong with version 2.1?

nikita-bedmutha:
@daniel-beck , yeah so we want to start with 1.0.0, which was our first version before hiding the plugin. Infact we will be pushing 1.0.1 changes soon after our product is live, and then that will be the latest version which we want hosted on jenkins update center. Should I also include qualys-api-secuirity-1.0.0 in this PR then? So that whenever we push new changes and perform mvn prepare release, we will see 1.0.1 directly available with JUC. What is your suggestion for this?

daniel-beck:
As this is a new plugin I would be disinclined to suspend an accidentally high (but not crazy) version number. I'll wait for input from others.

nikita-bedmutha:
@daniel-beck , we were actually experimenting automation for our releases, we actually have many (3 )plugins already published - 
https://github.com/jenkinsci/qualys-cs-plugin
https://github.com/jenkinsci/qualys-was-plugin
https://github.com/jenkinsci/qualys-vm-plugin

So everytime manually publishing the plugin was getting hectic, so we tried automation and since qualys-api-security was not in production yet, we tried experiments with it with diff version increments and thats how ended up with high number.

PR #452, opened 2020-09-28
Qualys module ready for release, so make plugin visible again.

nikita-bedmutha:


nikita-bedmutha:
@daniel-beck, can you please have a look at this PR? We actually need to release the plugin as soon as possible.

daniel-beck:
@nikita-bedmutha To clarify, you've abandoned https://github.com/jenkins-infra/update-center2/pull/450 and are fine with the versions you created? Once it's published and users install it, there's no way back.

nikita-bedmutha:
@daniel-beck , yes, we are fine with using current version and we will build on top of it.

PR #453, opened 2020-10-02
chore: add diagrams showing update processes

garethjevans:


MarkEWaite (approved):
That looks great!  Thanks very much

site/README.md
```diff
@@ -10,6 +10,13 @@ The service this website provides is as follows:
  1. We use the version number to redirect the traffic to the right update site,
     among all the ones that we generate.
 
+### What happens when 'Update Now' is pressed?
```
daniel-beck, line 13: Third level header directly below a first level header makes no sense.

Beyond formatting, this is also unrelated to anything else going on in this document.

daniel-beck (changes requested):
Regarding the chosen format: Given how trivial these flows are, please provide a textual version. Even if you did provide the source files, this is much more cumbersome to update, impossible to search, and less accessible, as image files.

Beyond that, the location you add these does not make sense. Granted, the docs in this repo are a bit distributed, but there are better locations: Perhaps some of this can be integrated into https://github.com/jenkins-infra/update-center2/blob/master/site/LAYOUT.md#top-level-htaccess-file (for the mirror redirection stuff) and https://github.com/jenkins-infra/update-center2/blob/master/site/LAYOUT.md#known-consumers expanded to explain how Jenkins consumes this.

In general though, I'm not a big fan of putting documentation outside the context it applies to; so exactly where mirrorbits redirects to, or even what sort of mirror infra exists, has no real reason to be in this repo. It's just going to get stale.

garethjevans:
abandoning

PR #459, opened 2020-10-08
Suspend persona (SECURITY-2046) and nerrvana (SECURITY-2097)

daniel-beck:
Forgot to include these in the warnings update 😅


daniel-beck:
Looks like CI is down, so merging without.

PR #461, opened 2020-10-21
Add support for getting default branch name, and issues urls

halkeye:


halkeye:
superseeds #355

daniel-beck:
Overall looks reasonable. I added some minor changes.

One thing I was wondering about: Why are we claiming a default branch of `master` when we don't even publish the SCM? All the early returns in `HPI#getDefaultBranch` should return `null`, as well as some in `GitHubSource#getDefaultBranch(GitHubRepo)`. I have it prepared locally but not yet pushed in case I'm missing something. Am I?

daniel-beck:
Also added the ability to override the issue tracker URL if needed. Some plugin have weird plugin IDs that aren't reflected in component names.

timja (approved):

halkeye:
> Why are we claiming a default branch of `master` when we don't even publish the SCM?

I think i was trying to avoid breaking things by keeping old behavior, but your absolutely right, its a new field, and plugin site shouldn't be trying to get readme if there's no repo data anyways.

daniel-beck:
> I have it prepared locally but not yet pushed

Thanks 😕

daniel-beck (approved):
I think this is good to go now 👍

olblak (approved):

yJunS (approved):

SonyViao (approved):

daniel-beck:
After some more discussions, we've decided to store the metadata in plugin metadata ( https://github.com/jenkins-infra/repository-permissions-updater ) and only consume the generated `issues.index.json` here, transforming output a bit. So we won't need the overrides and such.

https://github.com/jenkins-infra/repository-permissions-updater/pull/1806 implemented this and has further references.

daniel-beck:
#492 implements the issue tracker metadata information now as pass-through from https://github.com/jenkins-infra/repository-permissions-updater.

@halkeye I'll add default branches to the JSON in a separate PR if you still want/need that for the plugin site (to render docs from a non-`master` default branch IIRC), just let me know.

halkeye:
yes please!

daniel-beck:
#495 takes care that. Meanwhile I'm closing this.

PR #462, opened 2020-10-23
[INFRA-2487] Mark brakeman plugin as deprecated

daniel-beck:
This will add a deprecation warning to https://plugins.jenkins.io/brakeman/ (and inside Jenkins) similar to https://plugins.jenkins.io/analysis-core/

It's a bit redundant, given the dependency, but cannot hurt. It also links to the issue that explains what users can do instead now (and will continue to show up even once distribution has been suspended for INFRA-2487).

@presidentbeef @uhafner


daniel-beck:
🤷 Someone will complain if this is wrong.

PR #463, opened 2020-10-25
Ignore spotinst version 2.0.26

adivos:
Ignore spotinst version 2.0.26
This release update node details and by that causing Jenkins to reset its computer


daniel-beck:
> Last released: 3 weeks ago

There's a high chance many instances already have it installed, so it's probably a good idea to work on a regression fix ASAP and release 2.0.27. At a minimum, perhaps the specific responsible change could be reverted, unless that would cause new problems?

adivos:
we are currently testing a fix and will be released in the next couple of days

PR #466, opened 2020-11-04
[INFRA-2487] Suspend analysis-core and all dependent plugins

daniel-beck:
This will remove the plugins from update sites, and all instances with one of these plugins installed will show a deprecation message linking to INFRA-2487.

@uhafner @MarkEWaite @oleg-nenashev


timja (approved):

uhafner:
:wave:

PR #467, opened 2020-11-08
Avoid labels used by a single plugin

zbynek:
These labels don't make it easier to browse the update center because they are only assigned to a single plugin; `plugin-misc` was used twice but it seems to be an older syntax for commonly used `misc`.


daniel-beck (approved):
All of these are from https://github.com/jenkins-infra/update-center2/pull/316 which probably just imported what was defined, so this makes sense.

timja (approved):

MarkEWaite (approved):

yJunS (approved):

resources/allowed-github-topics.properties
```diff
@@ -60,10 +57,6 @@ page-decorator
 parameter
 performance
 pipeline
-pipeline-graph-analysis
-plugin-external
-plugin-misc
-plugin-post-build
 plugin-test
```
daniel-beck, line 60:
```suggestion
plugin-test
```

Only one plugin has this label AFAICT, should just be `test`?

zbynek: Good 👀 , fixed in last commit.

daniel-beck (approved):

PR #456, opened 2020-10-06
Suspend distribution of qualys-api-security-2.1.0 due to ambiguous version

daniel-beck:
Reference: https://github.com/jenkinsci/qualys-api-security-plugin/commits/master

I really need to get #412 over the finish line.

FYI @nikita-bedmutha please re-release 2.1.0 as 2.1.1.

PR #412, opened 2020-08-01
Remove ambiguously versioned releases

daniel-beck:
lib-versionnumber considers versions `whatever` and `whatever.0` to be equivalent. This means update-center2 cannot properly determine which of the two to consider 'newer' and advertise (assuming no other differences).

A similar problem can occur with a groupId change, `org.acme:my-plugin:1.0` and `com:acme:my-plugin:1.0` can both exist and one will take precedence over the other.

Right now, we choose one over the other essentially randomly. It appeared that the dependency updates in 3.3.1 compared to 3.3 changed something, which caused me to file https://github.com/jenkins-infra/update-center2/pull/411.

This change would result in _neither_ version to be published. Basically, creating both `X` and `X.0` would suspend distribution of both.

Thoughts?

----

The current implementation appears to have the weakness that having `X`, `X.0`, and `X.0.0` is expected to randomly admit one of them. I'll fix it if we get consensus that doing this at all isn't a terrible idea, I'll work on this some more.

PR #468, opened 2020-11-09
Keep older ambiguous releases

daniel-beck:
This builds on top of, and is an alternative to, #412.

The difference is that artifacts from Artifactory have a timestamp, and this implementation will advertise the _older_ release.

So, if 1.2 already exists, and 1.2.0 is released afterwards, 1.2 will keep being advertised.

This should be less disruptive than the "erasing" behavior of #412, which would end up removing perfectly good releases that may have existed for a long time (e.g. in #456 the previous release existed for half a year).

Thoughts?


timja (approved):

oleg-nenashev (approved):
I am fine with both this and the previous pull request

yJunS (approved):

oleg-nenashev (approved):
Re-+1

PR #469, opened 2020-11-14
Use template for download pages

zbynek:
Couple of things are still TODO, like the opengraph image, also the template should have a placeholder for `{{ content }}` to make substitutions easier.

image

@daniel-beck is that roughly what you had in mind?


daniel-beck:
What am I supposed to do now 😭

oleg-nenashev:
Looks pretty good, thanks @zbynek !

src/main/java/io/jenkins/update_center/IndexHtmlBuilder.java
```diff
@@ -44,36 +53,56 @@
...
+        }
+        template = globalTemplate.replace("{{ title }}", title)
+                .replace("{{ description }}", "Download previous versions of " + title);
+        content = new StringBuilder("<div id='grid-box'><div class=\"container\"><h1 class=\"mt-3\">")
+                .append(title).append("</h1><ul class=\"artifact-list\">\n");
+    }
+
+    private void initTemplate() {
+        Request request = new Request.Builder()
+                .url("https://www.jenkins.io/template/").get().build();
```
halkeye, line 88: so update center is supposed to be not jenkins specific right? I think it should not try and download a new template every time, but instead be something saved without the top bar.

Also is this just a single index.html? or one per directory? could it call /template a bunch of times per directory?

daniel-beck: One of the reasons this isn't merged yet 😉 @zbynek was just _too_ helpful after I asked for the design, I didn't get a chance to explain the implementation/configuration side of things.

zbynek: It's a lazily initiated static field, so it should be downloaded at most once per start of the Java process and reused for all the plugins. AFAIK the [bash script](https://github.com/jenkins-infra/update-center2/blob/master/site/generate.sh) starts the Jar twice, so performance should not be an issue.

>  implementation/configuration side of things

so what is the expectation here? :)

daniel-beck: > AFAIK the bash script starts the Jar only twice, so performance should not be an issue.

Yes, once to determine tiers, once to generate them all.

> so what is the expectation here? :)

No expectations for you, but I'm happy to accept your help 😃 My plan was sort of the following:

* New command line argument that configures the template, with the previous behavior (mostly) remaining the default.
* Add that argument to the standard invocation in the wrapper shell script so it gets applied.
* Implement template switching e.g. using an enum/strategy pattern sort of thing perhaps? Unsure in advance. Should be pretty easy to make it nicer as it is today in any case. Possibly split latest links and releases pages creation up if they don't share enough structure anymore.
* Consider moving the previous output into a "template" resource file to largely mirror the actual template file from jenkins.io except have it locally.
* Look into moving the CSS into a new HAML-based template on jenkins.io in the `templates/` folder.
* In general this is an opportunity to clean up the `IndexHtmlBuilder`. Figure out a nicer way to stitch together the output, perhaps abstract specific elements away behind a more semantic API. I basically left this class alone while I rewrote everything else earlier this year, so this is a nice opportunity to make this suck less.

(Of course some of these are really easy to defer until later, especially the cleanup oriented ones.)

Regarding template design, I think the proposed design is a huge improvement over what we currently have! A few suggestions:

* Make sure latest links (from `LatestLinkBuilder`) looks reasonable as well. Untested, but also goes through this code so may be impacted in unexpected ways.
* Use actual labels instead of (or in addition to) artifact ID. "Matrix Authorization Strategy", not (only) "matrix-auth"
* In the current design, it's a bit difficult to read the version number with the date immediately following, these should be more separated.
* Link to the plugin's plugin site page, if it's a plugin.

I played around with your design a bit, and this is what I came up with:

Screenshot

(Yes, I got my own plugin's name wrong 😄)

<details>
<summary>quick and dirty code</summary>

```
<style>.artifact-list {
  padding: 0;
}
.artifact-list .title {
  background-image: url(https://www.jenkins.io/images/jar.png);
  background-position: left center;
  background-repeat: no-repeat;
  list-style: none;
  padding-left: 2rem;
  margin-bottom: .5rem;
}
.container h2 {
    margin-top: 1rem;
}
.artifact-list td {
    padding-bottom: 0.5rem;
}
.artifact-list .title {
    font-size: 1.5rem;
  padding-right: 1rem;
}
.artifact-list .metadata {
    font-size: 0.8rem;
    color: #666;
}
</style>


<div id='grid-box'><div class="container">
    <h1 class="mt-3">Matrix Authorization Plugin</h1>
    <h3><span style="font-family: monospace;">matrix-auth</span></h3>
    <a href="https://plugins.jenkins.io/matrix-auth/">View on plugin site</a>
    <h2>Releases</h2>
    <table class="artifact-list">
        <tr>
            <td class="title" colspan="2">
                <a href='/latest/matrix-auth.hpi'>
                    Permanent to Latest Release
                </a>
            </td>
        </tr>
        <tr>
            <td class="title">
                <a href='/download/plugins/matrix-auth/2.5.1/matrix-auth.hpi'>
                    2.5.1
                </a>
            </td>
            <td class="metadata">
                    <div class="released">Released: Jun 29, 2020</div>
                    <div class="checksums">SHA-1: <code>f6046479f0a0a84de0738ac3fa144879de0f8598</code></div>
                    <div class="checksums">SHA-256: <code>7299ff13f61d4a84ccc65cea50f082bd256fd35657e16dfab081666781292c40</code></div>
            </td>
        </tr>
        <tr>
            <td class="title">
                <a href='/download/plugins/matrix-auth/2.5/matrix-auth.hpi'>
                    2.5
                </a>
            </td>
            <td class="metadata">
                    <div class="released">Released: Oct 14, 2019</div>
                    <div class="checksums">SHA-1: <code>c0d297bc809716fd3fe31427aee1391c88a59374</code></div>
                    <div class="checksums">SHA-256: <code>db99c3ecb22931ad49a81f7209c4187f2ad867f5d09ec4b4e7d19a915f543d63</code></div>
            </td>
        </tr>
    </table>
</div></div>
```

</details>

zbynek: I'd like to help with this, I'll try to follow the plan you outlined. Btw. is there a way to generate the latest links without downloading 30GB of old Jenkins versions?

zbynek:
@daniel-beck I updated the design to be more similar to your version, but it's using a list rather than a table. Please check if it looks OK. Once the design is release ready I can push it to jenkins.io repo and then continue with this PR.

zbynek:
@daniel-beck will you have time to check this? (not urgent)

daniel-beck:
@zbynek Still on my list, unfortunately my spare time was limited recently and I first wanted to unblock Gavin's work on issue tracker metadata. Sorry it's taking so long!

zbynek:
No problem, the issue tracker is definitely more important.

daniel-beck:
Had a first look yesterday and this looks great so far. 👍

Planning to take a more in-depth look later this week.

src/main/java/io/jenkins/update_center/IndexHtmlBuilder.java
```diff
@@ -34,8 +37,17 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
+import java.util.List;
+import java.util.Objects;
+import java.util.logging.Level;
+import java.util.logging.Logger;
```
daniel-beck, line 50: There's a bunch of unused imports in this file.

daniel-beck (approved):
This looks fantastic. I'd be happy to merge this as is. Thanks a lot for taking care of this!

src/main/java/io/jenkins/update_center/IndexHtmlBuilder.java
zbynek: Imports fixed. Also now it's using the template from jenkins.io so the code could be simplified a bit.

daniel-beck: > should I merge master to this or rebase?

Either works, I'd probably squash anyway. Thanks for checking!

daniel-beck (approved):
Very nice, thanks!

zbynek:
@daniel-beck when do you plan to merge this?

daniel-beck:
@zbynek When I have a chance to release and then monitor the output; probably this weekend.

daniel-beck:
Quick release update: trusted.ci (the system we generate update sites on) has problems at the moment, builds usually take >1 hr instead of 3 minutes since at least Friday. That's not an infra situation I want to deploy changes into, so I'm holding this off for a bit longer.

zbynek:
@daniel-beck thanks for the update. What makes the builds so slow?

daniel-beck:
@zbynek My guess is unintended side effects of changes to agents. No response so far on IRC, but between Friday being a CloudBees holiday and today being Whit Monday in large parts of the EU, I'm not surprised.

daniel-beck:
`trusted.ci` is down again… after it was down last Sunday as well…

daniel-beck:
> @daniel-beck I updated the design to be more similar to your version, but it's using a list rather than a table. Please check if it looks OK. Once the design is release ready I can push it to jenkins.io repo and then continue with this PR.

We're currently running off [a branch with these changes](https://github.com/jenkins-infra/update-center2/tree/download-pages), and I noticed the layout kind of breaking probably because of this (it didn't in my experiment), and not just for unusual version numbers, but also the fairly common 2.0 / 2.0.1 / 2.0.1.1 format: https://updates.jenkins.io/download/plugins/matrix-auth/

https://updates.jenkins.io/download/plugins/platformlabeler/ also shows that the choice on non-monospace font meant that JEP-229 version numbers will necessarily make the page look broken – at least that would be really straightforward to fix.

I'd be happy to merge as in for this to make it out of review hell if you want, but ideally we'd have a quick followup to address this. (Alternatively, explaining why I'm wrong is always an option).

zbynek:
@daniel-beck thanks for testing. I've made some changes in CSS to make this work for all screen sizes (big screen: table layout, small screen: list layout). The change will only affect locally generated files, if you think it's OK I'll do the same CSS change in the jenkins.io repo.

daniel-beck:
@zbynek Looks great! There's a bit of weird behavior between 600-770 px wide (the SHA-256 line is split into two lines), but that's nitpicking now (and might not apply to the 'real' template).

IIUC this is good to merge now, since any changes would only be in the template?

zbynek:
@daniel-beck yes, this one is good to go. I'll check the breakpoint for the upcoming PR in jenkins.io.

PR #470, opened 2020-11-28
[INFRA-2754] Cache tool download

daniel-beck:
No need to download this every few minutes.


oleg-nenashev (approved):
IIUC the production deployment build will still use a new file each time, right?

daniel-beck:
> IIUC the production deployment build will still use a new file each time, right?

No, since it's running on the same agent every time, otherwise the heavy caching of everything else wouldn't work.

oleg-nenashev:
OK. Whatever works then

PR #474, opened 2020-12-09
SECURITY-157 is fixed in mask-passwords 2.13.0

ineffyble:
As per the [release notes](https://github.com/jenkinsci/mask-passwords-plugin/releases/tag/mask-passwords-2.13) this was fixed back in April, but the warning was never updated.


daniel-beck:
Thanks for filing this @ineffyble I'll review it and merge once confirmed.

ineffyble:
@daniel-beck I think the regex is right, it passed tests against all the release numbers I could see on the plugin page, but a second pair of eyes is appreciated.

daniel-beck:
> I could see on the plugin page

It also needs to match all the release numbers you cannot see 😁 Don't worry, I'll take a look and fix if needed.

resources/warnings.json
```diff
@@ -5469,7 +5469,7 @@
     "versions": [
       {
         "lastVersion": "2.12.0",
-        "pattern": ".*"
+        "pattern": "(2[.]1[0-2]|2[.][0-9])([.-].+|)"
```
daniel-beck, line 5472:
```suggestion
        "pattern": "(1|2[.]1[0-2]|2[.][0-9])([.-].+|)"
```

Version 1.0 existed but is not currently published, better safe than sorry.

PR #476, opened 2020-12-11
Add support for snapshot versions

daniel-beck:
https://github.com/jenkinsci/jenkins/pull/5106#issuecomment-743093376 points out that redirects don't work for snapshot versions (any more?).


timja (approved):
woot thanks

olblak (approved):

daniel-beck:
On second thought, we should go one version lower (i.e. treat 1.2.3-SNAPSHOT like 1.2.2), but 🤷 I doubt this makes a difference very often.

PR #480, opened 2021-01-16
Suspend distribution of github-api 1.222

daniel-beck:
@bitwiseman 

We have no good way to inform users they have a bad release installed. The usual resolution is to publish a new version. Versions that have been used are burned afterwards. My general advice in such cases is to accept the mistake and move on with the new version numbers; it's simply unfortunate that the versioning scheme used here makes that awkward.

Unsure what the best way forward is; if it is to suspend 1.222, it should be done ASAP.


bitwiseman:
@daniel-beck 
I see.  Thank you. I will do a version jump to 1.223 for my next release.

bitwiseman:
But this looks good to me. 
Thanks for doing this.

daniel-beck:
As discussed with @bitwiseman elsewhere:

We'll be able to see the scope of the problem in the January usage stats. We can always unsuspend / do a huge version bump later, but continuing to distribute 1.222 would just make the problem worse.

PR #481, opened 2021-01-19
Remove unnecessary overrides

zbynek:
These plugins have meanwhile released a version with either the exact same URL or a better one (moving docs to GitHub).

### DIFF
--- a/resources/wiki-overrides.properties
+++ b/resources/wiki-overrides.properties
@@ -80,6 +80,3 @@ handlebars=https://github.com/jenkinsci/js-libs/blob/master/handlebars/README.md
 jquery-detached=https://github.com/jenkinsci/js-libs/blob/master/jquery-detached/README.md
 momentjs=https://github.com/jenkinsci/js-libs/blob/master/momentjs/README.md
 numeraljs=https://github.com/jenkinsci/js-libs/blob/master/numeraljs/README.md
-
-# Required until a newer version than 1.0.0 is released
-synopsys-polaris=https://github.com/jenkinsci/synopsys-polaris-plugin
--- a/resources/wiki-overrides.properties
+++ b/resources/wiki-overrides.properties
@@ -4,9 +4,6 @@
 # Required until a newer version than 1.1 is released
 archived-artifact-url-viewer=https://wiki.jenkins-ci.org/display/JENKINS/Archived+Artifact+Url+Viewer+PlugIn
 
-# Required until a newer version than 0.3 is released
-plugin-usage-plugin=https://wiki.jenkins-ci.org/display/JENKINS/Plugin+Usage+Plugin+(Community)
-
 # Required until a newer version than 1.1.0 is released
 skype-notifier=http://wiki.jenkins-ci.org/display/JENKINS/Skype+Plugin
 
@@ -17,19 +14,16 @@ chef-tracking=https://wiki.jenkins-ci.org/display/JENKINS/Chef+Tracking+Plugin
 chosen-views-tabbar=https://wiki.jenkins-ci.org/display/JENKINS/Chosen+Views+Tab+Bar
 ## Source seems to *not* be under jenkinsci org: https://github.com/ThoughtsOnMobile/cocoapods-jenkins-integration
 cocoapods-integration=https://wiki.jenkins-ci.org/display/JENKINS/CocoaPods+Plugin
-cucumber-reports=https://wiki.jenkins-ci.org/display/JENKINS/Cucumber+Reports+Plugin
 custom-view-tabs=https://wiki.jenkins-ci.org/display/JENKINS/Custom+View+Tabs+Plugin
 ## Currently has the <url> tag, BUT with a typo https://wiki.jenkins-ci.org/display/JENKINS/Deploment+Notification+Plugin
 deployment-notification=https://wiki.jenkins-ci.org/display/JENKINS/Deployment+Notification+Plugin
 database-drizzle=https://wiki.jenkins-ci.org/display/JENKINS/Drizzle(+MySQL)%20Database%20Plugin
 diskcheck=https://wiki.jenkins-ci.org/display/JENKINS/DiskCheck+Plugin
-docker-build-publish=https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Docker+Build+and+Publish+plugin
 dynamicparameter=https://wiki.jenkins-ci.org/display/JENKINS/Dynamic+Parameter+Plug-in
 figlet-buildstep=https://wiki.jenkins-ci.org/display/JENKINS/Figlet+plugin
 groovyaxis=https://wiki.jenkins-ci.org/display/JENKINS/GroovyAxis
 hckrnews=https://wiki.jenkins-ci.org/display/JENKINS/Hckrnews+Plugin
 icon-shim=https://wiki.jenkins-ci.org/display/JENKINS/Icon+Shim+Plugin
-image-gallery=https://wiki.jenkins-ci.org/display/JENKINS/Image+Gallery+Plugin
 jenkins-testswarm-plugin=https://wiki.jenkins-ci.org/display/JENKINS/testswarm-plugin
 job-direct-mail=https://wiki.jenkins-ci.org/display/JENKINS/Job+Direct+Mail+Plugin
 jython=https://wiki.jenkins-ci.org/display/JENKINS/Jython+Plugin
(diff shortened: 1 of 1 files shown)


daniel-beck:
@zbynek Thanks! I'll take a look as soon as I can.

olblak (approved):
I checked that all those plugins had documentation on plugins.jenkins.io but are they supposed to also have the URL specified in the pom.xml? At least I identify that the serenity doesn't.

resources/wiki-overrides.properties
```diff
@@ -4,9 +4,6 @@
 # Required until a newer version than 1.1 is released
 archived-artifact-url-viewer=https://wiki.jenkins-ci.org/display/JENKINS/Archived+Artifact+Url+Viewer+PlugIn
 
-# Required until a newer version than 0.3 is released
-plugin-usage-plugin=https://wiki.jenkins-ci.org/display/JENKINS/Plugin+Usage+Plugin+(Community)
```
daniel-beck, line 8: https://github.com/jenkinsci/plugin-usage-plugin/commit/c135ff6c5ca1d4f46bd1c6f576a70c9059bdc2b3 references the repo now. Changelog was migrated to a different file, the rest is identical.

resources/wiki-overrides.properties
```diff
@@ -17,19 +14,16 @@ chef-tracking=https://wiki.jenkins-ci.org/display/JENKINS/Chef+Tracking+Plugin
 chosen-views-tabbar=https://wiki.jenkins-ci.org/display/JENKINS/Chosen+Views+Tab+Bar
 ## Source seems to *not* be under jenkinsci org: https://github.com/ThoughtsOnMobile/cocoapods-jenkins-integration
 cocoapods-integration=https://wiki.jenkins-ci.org/display/JENKINS/CocoaPods+Plugin
-cucumber-reports=https://wiki.jenkins-ci.org/display/JENKINS/Cucumber+Reports+Plugin
```
daniel-beck, line 20: https://github.com/jenkinsci/cucumber-reports-plugin/commit/50cb26da9e42955b98ee19f96896175e5b5ac05b#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R16 references the repo now, but the wiki page was just a placeholder, so this is a good improvement.

resources/wiki-overrides.properties
```diff
@@ -17,19 +14,16 @@ chef-tracking=https://wiki.jenkins-ci.org/display/JENKINS/Chef+Tracking+Plugin
...
 custom-view-tabs=https://wiki.jenkins-ci.org/display/JENKINS/Custom+View+Tabs+Plugin
 ## Currently has the <url> tag, BUT with a typo https://wiki.jenkins-ci.org/display/JENKINS/Deploment+Notification+Plugin
 deployment-notification=https://wiki.jenkins-ci.org/display/JENKINS/Deployment+Notification+Plugin
 database-drizzle=https://wiki.jenkins-ci.org/display/JENKINS/Drizzle(+MySQL)%20Database%20Plugin
 diskcheck=https://wiki.jenkins-ci.org/display/JENKINS/DiskCheck+Plugin
-docker-build-publish=https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Docker+Build+and+Publish+plugin
```
daniel-beck, line 26: https://github.com/jenkinsci/docker-build-publish-plugin/commit/0240b543c668767a85b20ba9b9ef839c696a2244 references the same URL.

resources/wiki-overrides.properties
```diff
@@ -17,19 +14,16 @@ chef-tracking=https://wiki.jenkins-ci.org/display/JENKINS/Chef+Tracking+Plugin
...
 dynamicparameter=https://wiki.jenkins-ci.org/display/JENKINS/Dynamic+Parameter+Plug-in
 figlet-buildstep=https://wiki.jenkins-ci.org/display/JENKINS/Figlet+plugin
 groovyaxis=https://wiki.jenkins-ci.org/display/JENKINS/GroovyAxis
 hckrnews=https://wiki.jenkins-ci.org/display/JENKINS/Hckrnews+Plugin
 icon-shim=https://wiki.jenkins-ci.org/display/JENKINS/Icon+Shim+Plugin
-image-gallery=https://wiki.jenkins-ci.org/display/JENKINS/Image+Gallery+Plugin
```
daniel-beck, line 32: https://github.com/jenkinsci/image-gallery-plugin/commit/58bd07766ead88e96b1261fbb234f070ed9dbfe7 references the repo now, but docs were migrated.

resources/wiki-overrides.properties
```diff
@@ -47,9 +41,6 @@ prereq-buildstep=http://wiki.jenkins-ci.org/display/JENKINS/Prerequisite+build+s
 project-health-report=https://wiki.jenkins-ci.org/display/JENKINS/Project+Health+Report+Plugin
 puppet=https://wiki.jenkins-ci.org/display/JENKINS/Puppet+Plugin
 ruby-runtime=https://wiki.jenkins-ci.org/display/JENKINS/Ruby+Runtime+Plugin
-serenity=https://wiki.jenkins-ci.org/display/JENKINS/Serenity+Plugin
```
daniel-beck, line 50: https://github.com/jenkinsci/serenity-plugin/commit/ef4c5b00f4638fece63b3f7e31640d6298b76818 references the repo now, but docs were migrated.

resources/wiki-overrides.properties
```diff
@@ -47,9 +41,6 @@ prereq-buildstep=http://wiki.jenkins-ci.org/display/JENKINS/Prerequisite+build+s
...
-## Currently has the <url> tag, BUT with a typo http://http://wiki.jenkins-ci.org/display/JENKINS/Skytap+Cloud+CI+Plugin
-skytap=https://wiki.jenkins-ci.org/display/JENKINS/Skytap+Cloud+CI+Plugin
```
daniel-beck, line 52: This is an interesting case because it used to be wrong to do this: https://github.com/jenkinsci/skytap-cloud-plugin/commit/c7bf2fe43991a0527d003e5e1e8ad7f02d5839ca

Now it's not, but the wiki page is still needed because the repo only contains placeholder docs — but redirects to the plugin site 👎  So this will make all of the documentation inaccessible.

resources/wiki-overrides.properties
```diff
@@ -62,13 +53,9 @@ vaddy-plugin=https://wiki.jenkins-ci.org/display/JENKINS/VAddy+Plugin
 violations=https://wiki.jenkins-ci.org/display/JENKINS/Violations
 xpdev=https://wiki.jenkins-ci.org/display/JENKINS/XP-Dev+Plugin
 
-# Workaround for INFRA-922
-absint-astree=https://wiki.jenkins-ci.org/pages/viewpage.action?pageId=99910545
```
daniel-beck, line 66: https://github.com/jenkinsci/absint-astree-plugin/commit/3a03a5b334325cba2018093ec77fed6ffd2bd8c2 references the same URL.

resources/wiki-overrides.properties
```diff
@@ -62,13 +53,9 @@ vaddy-plugin=https://wiki.jenkins-ci.org/display/JENKINS/VAddy+Plugin
...
-
 github-scm-trait-notification-context=https://wiki.jenkins.io/display/JENKINS/Github+Custom+Notification+Context+SCM+Behaviour+Plugin
 ca-apm=https://wiki.jenkins.io/display/JENKINS/CA+APM+Plugin+1.x
 comments-remover=https://wiki.jenkins.io/display/JENKINS/XComment.io+-+Comments+Remover+Plugin
-qtest=https://github.com/jenkinsci/qtest-plugin
```
daniel-beck, line 71: https://github.com/jenkinsci/qtest-plugin/commit/eb33a469790bd2fb31ac78889fbf3b3f6b1bb283 references the same URL.

daniel-beck (changes requested):
Looks like this will cause problems for `skytap`.

daniel-beck:
> I checked that all those plugins had documentation on plugins.jenkins.io

@olblak 
The URLs specified here control where plugins.jenkins.io gets the documentation from. So you confirmed that the current state is OK, not the change implemented here.

resources/wiki-overrides.properties
zbynek: There are two <url> tags within the POM and the one outside of <scm> takes precendence -- I think this should make no difference for Skytap

resources/wiki-overrides.properties
```diff
@@ -80,6 +80,3 @@ handlebars=https://github.com/jenkinsci/js-libs/blob/master/handlebars/README.md
 jquery-detached=https://github.com/jenkinsci/js-libs/blob/master/jquery-detached/README.md
 momentjs=https://github.com/jenkinsci/js-libs/blob/master/momentjs/README.md
 numeraljs=https://github.com/jenkinsci/js-libs/blob/master/numeraljs/README.md
-
-# Required until a newer version than 1.0.0 is released
-synopsys-polaris=https://github.com/jenkinsci/synopsys-polaris-plugin
```
zbynek, line 98: @daniel-beck I missed this one; 1.1.0 is released with this change https://github.com/jenkinsci/synopsys-polaris-plugin/blob/1.1.0/build.gradle#L28

resources/wiki-overrides.properties
daniel-beck: Oh right, I failed at reading XML. Sorry about that.

The URL was set in https://github.com/jenkinsci/skytap-cloud-plugin/commit/ccfe470652a023433c59d0b99ce36bb658a8cdb8 but is linking to plain HTTP. So the original problem doesn't exist anymore, but a different one.

daniel-beck: https://github.com/jenkins-infra/update-center2/blob/f0fff816539341fdfbf2ad206556f5f1dec12858/src/main/java/io/jenkins/update_center/HPI.java#L406 should take care of that though, so I'm fine with this change.

PR #484, opened 2021-02-23
Stop ignoring sshd

timja:
See https://github.com/jenkinsci/jenkins/pull/5049

https://github.com/jenkins-infra/update-center2/pull/364

cc @jglick

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -407,6 +407,9 @@ configuration-as-code-1.37
 kubesphere-extension-0.1.0
 kubesphere-extension-0.2.0
 
+# sshd module was packaged as hpi at least for a time
+sshd-1.2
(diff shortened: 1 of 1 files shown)


jglick (approved):

resources/artifact-ignores.properties
```diff
@@ -408,7 +408,7 @@ kubesphere-extension-0.1.0
 kubesphere-extension-0.2.0
 
 # sshd module was packaged as hpi at least for a time
-sshd
+sshd-1.2
```
daniel-beck, line 411: What about 1.0 and 1.1?
```suggestion
sshd-1.0
sshd-1.1
sshd-1.2
```

daniel-beck (changes requested):

PR #487, opened 2021-02-28
Suppress instance-identity plugin releases

jglick:
Pending https://github.com/jenkinsci/jenkins/pull/5304 @timja. Basically it is necessary to fully test the core patch _before_ releasing the plugin for the first time, using Incrementals.


timja (approved):

daniel-beck:
I would expect that currently, no real Jenkins instance is getting served an update site with this plugin offered…?

jglick:
> Is this just preparation for the release of 2.282 on Tuesday?

Whenever 2.282 comes, yes.

PR #491, opened 2021-03-11
Removing Liquibase Runner 1.4.7 due to CVE

r2liquibase:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=2020-2285

Thanks!

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -334,6 +334,7 @@ liquibase-runner-1.4.2
 liquibase-runner-1.4.3
 liquibase-runner-1.4.4
 liquibase-runner-1.4.5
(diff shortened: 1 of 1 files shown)


daniel-beck:
Could you provide some more context for this? We do not remove old versions just because they have vulnerabilities unless they're critical. Jenkins has built-in features that warn users about vulnerabilities in plugins they have installed.

r2liquibase:
Sure, @daniel-beck . Our security team would like to avoid users having the ability to download software that has a CVE. Furthermore, it is a VERY old release that has many, many issues that have since been fixed. Thanks!

r2liquibase:
Thanks!

PR #501, opened 2021-03-20
Return a tiny response for ?uctest requests

daniel-beck:
Log excerpt without this change:

```
(IP) - - [20/Mar/2021:20:03:42 +0000] "GET /update-center.json?uctest HTTP/1.1" 301 3555 "-" "Java/1.8.0_222"
(IP) - - [20/Mar/2021:20:03:42 +0000] "GET /current/update-center.json HTTP/1.1" 200 861701 "-" "Java/1.8.0_222"
```

With this change:
```
(IP) - - [20/Mar/2021:20:48:11 +0000] "GET /update-center.json?uctest HTTP/1.1" 200 3295 "-" "Java/1.8.0_222"

```

Clearly we don't need to send 860 KB for a connection check that ignores the returned content anyway: https://github.com/jenkinsci/jenkins/blob/56b6623915c50b4a2ef994a143a8fe0829587f3c/core/src/main/java/hudson/model/UpdateCenter.java#L1182-L1207

I think this gets requested whenever a file (core or plugin) is downloaded (updated/installed) for the first time since Jenkins started, so it's not for every plugin installation, but definitely too often to not add special handling.


timja (approved):
nice

daniel-beck:
This change is currently live as we're building from a branch right now:

https://updates.jenkins.io/update-center.json?uctest

PR #502, opened 2021-03-21
Use RPU permissions for 'developers' instead of Maven metadata

daniel-beck:
Downstream of https://github.com/jenkins-infra/repository-permissions-updater/pull/1880 and https://github.com/jenkins-infra/infra-reports/pull/27

PR currently fails because of infra not yet having the ID/display name maintainer mapping because trusted.ci is down, so a test assertion fails.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -28,7 +28,10 @@ The generator pulls information from:
   - default branch name
 * Jenkins usage statistics (see `Popularities.java`)
   - latest plugin installation numbers for `popularity` entries in update center JSON
-* Plugin issue tracker metadata JSON file on `reports.jenkins.io` (see `IssueTrackerSource.java`)
+* Various metadata JSON files on `reports.jenkins.io`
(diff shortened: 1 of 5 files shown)


oleg-nenashev (commented):
Should the old way be used as a fallback? It is still relevant when this repo is used to create custom update centers though it is not a very supported use case

daniel-beck:
> Should the old way be used as a fallback? It is still relevant when this repo is used to create custom update centers though it is not a very supported use case

Self-hosted update sites can provide bogus URL overrides and then not have maintainers information, so this tool still works. Jenkins does not consume this metadata at all anyway. It's for the old wiki macro (now the plugin site), so has no impact on self-hosted update sites. So I see no reason to keep that unused-by-us code around. See also the note at the top of [the readme](https://github.com/jenkins-infra/update-center2/blob/master/README.adoc).

oleg-nenashev (approved):

timja (approved):

PR #505, opened 2021-03-27
SECURITY-1559 was fixed in digitalocean-plugin-1.2.0

halkeye:


MarkEWaite (approved):
Confirmed the [release notes](https://github.com/jenkinsci/digitalocean-plugin/releases/tag/digitalocean-plugin-1.2.0) state that the issue is resolved.  Did not test the plugin release to prove it.

halkeye:
> Did not test the plugin release to prove it.

HOW DARE YOU

i mean thanks, I'm a little unsure i did the pattern right.

oleg-nenashev (changes requested):
The regexp wrong, because it won't match versions before 1.0, e.g. https://github.com/jenkinsci/digitalocean-plugin/releases/tag/digitalocean-plugin-0.16 . Fixing the first digit check is not enough, 2-digit version format needs to be supported as well

resources/warnings.json
```diff
@@ -7124,7 +7124,7 @@
     "versions": [
       {
         "lastVersion": "1.1",
-        "pattern": ".*"
+        "pattern": "1[.][01][.].+"
```
halkeye, line 7127: Since its a relatively small finite list, can i just cheat and list them out 
```suggestion
       "pattern": "^(0[.]1|0[.]10|0[.]11|0[.]12|0[.]13|0[.]14|0[.]14[.]1|0[.]15|0[.]16|0[.]17|0[.]18|0[.]19|0[.]19[.]1|0[.]19[.]2|0[.]2|0[.]3|0[.]4|0[.]4[.]1|0[.]5|0[.]6|0[.]6[.]1|0[.]7|0[.]8|0[.]9|1[.]0|1[.]1|1[.]1[.]1)$",
```

daniel-beck:
```suggestion
        "pattern": "(0|1[.][01])(|[.-].+)"
```

daniel-beck:
@halkeye Thanks. I'll confirm that the issue was fully resolved and then merge this PR.

PR #506, opened 2021-03-30
SECURITY-1928: fixed in release-plugin 2.11

Smasherr:
Thanks @daniel-beck for your support!


daniel-beck:
@Smasherr Thanks for submitting the PR. I will confirm the issue is fixed and then merge this.

resources/warnings.json
```diff
@@ -9130,7 +9130,7 @@
     "versions": [
       {
         "lastVersion": "2.10.2",
-        "pattern": ".*"
+        "pattern": "(1|2[.]\\d[0]?)(|[.-].+)"
```
daniel-beck, line 9133: The current regex will false identify future version 2.20 as affected.

```suggestion
        "pattern": "(1|2[.][0-9]|2[.]10[.]1)(|[.-].+)|(2[.]10)"
```

Smasherr: Good catch!

PR #508, opened 2021-04-07
Enable the observability GitHub topic

v1v:
### What

Enable the observability topic

### DIFF
--- a/resources/label-definitions.properties
+++ b/resources/label-definitions.properties
@@ -685,7 +685,6 @@ openshift-pipeline=pipeline scm-related post-build builder misc
 openshift-sync=cluster builder scm-related external
 openstack-cloud=cluster slaves
 openstack-heat=builder
-opentelemetry=observability administrative-monitor monitoring
 oslc-cm=notifier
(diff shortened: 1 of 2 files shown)


resources/label-definitions.properties
```diff
@@ -685,6 +685,7 @@ openshift-pipeline=pipeline scm-related post-build builder misc
 openshift-sync=cluster builder scm-related external
 openstack-cloud=cluster slaves
 openstack-heat=builder
+opentelemetry=observability administrative-monitor monitoring
```
oleg-nenashev, line 688: Better to use GitHub topics

daniel-beck: Agree, please use GH topics.

```suggestion
```

v1v: I thought it was required both! Thanks for the heads up :)

daniel-beck (approved):
Thanks! I'll merge this later today after I'm done with the security advisory.

oleg-nenashev (approved):

PR #509, opened 2021-04-09
Note fix of SECURITY-1510 in backlog plugin

daniel-beck:
Apparently this is fixed in version 2.5 of the plugin.

I'll confirm that and then merge this PR.


daniel-beck:
Confirmed fixed.

PR #510, opened 2021-04-12
Fix drecycler metadata by suspending it as it got renamed in 2014

halkeye:


resources/wiki-overrides.properties
```diff
@@ -80,3 +80,5 @@ handlebars=https://github.com/jenkinsci/js-libs/blob/master/handlebars/README.md
 jquery-detached=https://github.com/jenkinsci/js-libs/blob/master/jquery-detached/README.md
 momentjs=https://github.com/jenkinsci/js-libs/blob/master/momentjs/README.md
 numeraljs=https://github.com/jenkinsci/js-libs/blob/master/numeraljs/README.md
+drecycler=https://github.com/jenkinsci/diskcheck-plugin/blob/master/README.MD
```
daniel-beck, line 83: https://github.com/jenkinsci/diskcheck-plugin/commit/4361fff9713ef5e8079f36ca90663f599f6bd10f

We should just suspend distribution instead and advertise as deprecated, there's never going to be a new release of `drecycler`.

daniel-beck:
@halkeye This just adds a warning, suspensions are in https://github.com/jenkins-infra/update-center2/blob/master/resources/artifact-ignores.properties -- per commit message, this doesn't seem deliberate?

PR #512, opened 2021-04-14
Fix deploygate-plugin's docs

halkeye:


daniel-beck:
In theory a good idea, in practice there's no docs at the new URL anyway, and I doubt there will be, given the last release was 12 years ago.

PR #513, opened 2021-04-15
Switch to using wget with convert links to make sure index.html has absolute urls to the jenkins.io assets

halkeye:
Fixes the rendering of https://updates.jenkins.io so that it will look like the usual page rather than looking like this:

screencapture-updates-jenkins-io-2021-04-15-12_50_08-edit

### DIFF
--- a/site/generate.sh
+++ b/site/generate.sh
@@ -195,5 +195,5 @@ popd
 
 # copy other static resource files
 echo '{}' > "$WWW_ROOT_DIR/uctest.json"
-curl --location --fail https://www.jenkins.io/templates/updates/index.html > "$WWW_ROOT_DIR/index.html"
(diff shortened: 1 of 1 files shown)


halkeye:
@daniel-beck @MarkEWaite

timja (approved):

timja:
Ran the command locally seems fine

MarkEWaite (approved):
Are we confident that wget is available in the container where this runs?

Yes, there is a check in the script that causes the script to fail promptly if wget is not found on the PATH.

MarkEWaite:
Confirmed it works.  Thanks very much @halkeye !

halkeye:
Man I just came back to this PR 6 hours later and was totally shocked by the description and screenshots. Thanks @MarkEWaite 
I litterally said out loud "Did i make those screenshots? I don't remember doing that"

PR #520, opened 2021-05-08
Normalize issues URLs

zbynek:
This avoids redirects for the deprecation links and saves a few bytes of the JSON size.


oleg-nenashev (approved):
Thanks!

PR #522, opened 2021-05-13
Add deprecation links for suspended plugins

zbynek:
This should cover all suspended plugins with >200 (0.07%) known installs.

### DIFF
--- a/resources/deprecations.properties
+++ b/resources/deprecations.properties
@@ -59,6 +59,7 @@ codescan = https://git.io/JfaQb
 configuration-as-code-support = https://git.io/JfaHz
 # https://wiki.jenkins.io/display/JENKINS/CopyArchiver+Plugin
 copyarchiver = https://wiki.jenkins.io/x/ywJAAg
+copy-to-slave = https://jenkins.io/security/advisory/2018-03-26/
 cpptest = https://issues.jenkins.io/browse/INFRA-2487
 # https://wiki.jenkins.io/display/JENKINS/CppUnit+Plugin
 cppunit = https://wiki.jenkins.io/x/6oE5Ag
@@ -66,6 +67,7 @@ cppunit = https://wiki.jenkins.io/x/6oE5Ag
 dockerhub = https://git.io/JfaQF
 DotCi-Plugins-Starter-Pack = https://issues.jenkins.io/browse/INFRA-2487
 dry = https://issues.jenkins.io/browse/INFRA-2487
+dynamicparameter = https://jenkins.io/security/advisory/2017-04-10/
 # https://wiki.jenkins.io/display/JENKINS/Emotional+Hudson+Plugin
 emotional-hudson = https://wiki.jenkins.io/x/C4DX
 # https://github.com/jenkins-infra/update-center2/pull/257
@@ -84,6 +86,7 @@ gitorious = https://wiki.jenkins.io/x/ngDiAw
 google-desktop-gadget = https://wiki.jenkins.io/x/E4A3AQ
 # https://wiki.jenkins.io/display/JENKINS/Google+Code+Plugin
 googlecode = https://wiki.jenkins.io/x/DQC7
+grails = https://jenkins.io/security/advisory/2017-04-10/
 # https://wiki.jenkins.io/display/JENKINS/Hall+Plugin
 hall-jenkins = https://wiki.jenkins.io/x/qQghB
 # https://wiki.jenkins.io/display/JENKINS/HockeyApp+Plugin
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):
Thanks very much!

timja (approved):

resources/deprecations.properties
```diff
@@ -152,9 +167,11 @@ setenv = https://wiki.jenkins.io/x/YAtSAg
 sonargraph-plugin = https://git.io/JJ5h1
 # https://wiki.jenkins.io/pages/viewpage.action?pageId=63930505
 sonatype-ci = https://wiki.jenkins.io/x/iYDPAw
+svn-tag = https://jenkins.io/security/advisory/2017-04-10/
 swamp = https://issues.jenkins.io/browse/INFRA-2487
 tasks = https://issues.jenkins.io/browse/INFRA-2487
 # https://wiki.jenkins.io/display/JENKINS/TEPCO+Plugin
+tfs = https://issues.jenkins.io/browse/INFRA-2751
```
halkeye, line 174: is it not better to link to the [advisory](https://www.jenkins.io/security/advisory/2021-03-30/)? the issue... well the reason for suspension could easily be lost in history if someone else comments.

halkeye:
This is super impressive, i feel like it was a bunch of grinding to get this data.

resources/deprecations.properties
MarkEWaite: There are two advisories for TFS.  The INFRA ticket can be updated and contains more history.  I think we should continue with INFRA

daniel-beck: The security issues are not the reason for its suspension, so that would be wrong.

resources/deprecations.properties
```diff
@@ -168,5 +185,7 @@ vessel = https://groups.google.com/d/msg/jenkinsci-dev/L34eAMMWA5o/-AxtRdGsAAAJ
 warnings = https://issues.jenkins.io/browse/INFRA-2487
 # https://github.com/jenkinsci/xltestview-plugin/commit/ea034f9929b520e63b9ce15aed9bdb62354146cf
 xltest-plugin = https://git.io/JfaQK
+xtrigger = https://jenkins.io/security/advisory/2017-04-10/
```
daniel-beck, line 188: https://www.jenkins.io/blog/2017/04/10/security-advisory/#distributing-vulnerable-plugins or https://wiki.jenkins.io/display/JENKINS/Script+Security+Support+in+Plugins is probably a better reference for these plugins?

resources/deprecations.properties
timja: I thought the licensing issue was resolved? So what’s the reason still? Just no maintainer?

daniel-beck:
I wonder whether it would be useful to keep a reference of suspended plugins and their reasons around, e.g. on https://www.jenkins.io/security/plugins/ -- thoughts? Both would be more explicit than what we currently have (it seems wrong to edit a 4 year old blog post to add a list)

zbynek:
@daniel-beck sounds good. Should we set up the page first, then update this PR? Or do this PR as a first step and clean up the deprecation links once the security page is ready?

daniel-beck:
> @daniel-beck sounds good. Should we set up the page first, then update this PR? Or do this PR as a first step and clean up the deprecation links once the security page is ready?

Depends on whether you feel like waiting a bit longer, either works for me. I'd be happy to write the jenkins.io content later today or worst case tomorrow.

zbynek:
OK, then I'll wait for the security page.

daniel-beck:
Proposed in https://github.com/jenkins-infra/jenkins.io/pull/4359

The reference URL would be https://www.jenkins.io/security/plugins/#suspensions for all of them, I think that's good enough.

resources/deprecations.properties
```diff
@@ -125,7 +135,12 @@ origo-issue-notifier = https://wiki.jenkins.io/x/qgabAg
 paranoia = https://git.io/JfaQS
 # https://plugins.jenkins.io/perfectomobile/
 perfectomobile = https://developers.perfectomobile.com/display/PD/Legacy+%7C+Jenkins+plugin
+perforce = https://jenkins.io/security/advisory/2018-03-26/
```
daniel-beck, line 138: Re https://github.com/jenkins-infra/jenkins.io/pull/4359#issuecomment-841615950,

```suggestion
perforce = https://www.jenkins.io/security/advisory/2018-03-26/#SECURITY-373
```

Looks like a reasonable(ish) reference in the description there.

resources/deprecations.properties
```diff
@@ -125,7 +135,12 @@ origo-issue-notifier = https://wiki.jenkins.io/x/qgabAg
...
+persona = https://www.jenkins.io/security/plugins/#suspensions
 php = https://issues.jenkins.io/browse/INFRA-2487
+# https://github.com/jenkins-infra/update-center2/pull/384
+play-autotest-plugin = https://git.io/JsmWA
+pipeline-classpath = https://jenkins.io/security/advisory/2017-03-20
```
daniel-beck, line 143: Updated the suspensions list, so now:

```suggestion
pipeline-classpath = https://www.jenkins.io/security/plugins/#suspensions
```

resources/deprecations.properties
daniel-beck: Nobody stepping up to maintain it 🤷

zbynek:
@daniel-beck the jenkins.io PR is merged and online, can this one be merged as well?

daniel-beck:
@zbynek Sure, thanks for the ping!

PR #524, opened 2021-05-21
Spotbugs threshold to medium

zbynek:
* use try with resources as much as possible to avoid resource leaks
* add null pointer related annotations


pom.xml
```diff
@@ -122,6 +121,12 @@
           <redirectTestOutputToFile>true</redirectTestOutputToFile>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>com.github.spotbugs</groupId>
+        <artifactId>spotbugs-maven-plugin</artifactId>
+        <version>4.2.0</version>
+        <configuration><excludeFilterFile>${project.basedir}/spotbugs.xml</excludeFilterFile></configuration>
```
daniel-beck, line 128: Would it make sense to rename the file?
```suggestion
        <configuration><excludeFilterFile>${project.basedir}/spotbugs-excludes.xml</excludeFilterFile></configuration>
```
I was reading the diff from bottom to top and was confused why the filter rules listed seemed to be only for rules that don't make sense (and the comments indicated I'm correct), before I saw `excludeFilterFile` here. I have little experience with FB config though, maybe it's clear to everyone else.

src/main/java/io/jenkins/update_center/HPI.java
```diff
@@ -189,8 +194,8 @@ public String getDescription() throws IOException {
             ArtifactCoordinates coordinates = new ArtifactCoordinates(artifact.groupId, artifact.artifactId, artifact.version, "jar");
             try (InputStream is = repository.getZipFileEntry(new MavenArtifact(repository, coordinates), "index.jelly")) {
                 StringBuilder b = new StringBuilder();
-                HtmlStreamRenderer renderer = HtmlStreamRenderer.create(b, Throwable::printStackTrace, html -> LOGGER.log(Level.INFO, "Bad HTML: '" + html + "' in " + artifact.getGav()));
-                HtmlSanitizer.sanitize(IOUtils.toString(is, StandardCharsets.UTF_8), HTML_POLICY.apply(renderer), PRE_PROCESSOR);
+                @NonNull HtmlStreamRenderer renderer = HtmlStreamRenderer.create(b, Throwable::printStackTrace, html -> LOGGER.log(Level.INFO, "Bad HTML: '" + html + "' in " + artifact.getGav()));
```
daniel-beck, line 197: This looks weird, what's happening here?

src/main/java/io/jenkins/update_center/IndexTemplateProvider.java
```diff
@@ -18,6 +18,7 @@ public IndexHtmlBuilder newIndexHtmlBuilder(File dir, String title) throws IOExc
     }
 
     protected String initTemplate() {
+        assert Main.resourcesDir != null;
```
daniel-beck, line 21: Could we annotate it as such?

daniel-beck:
Overall looks great, just a few comments/questions above. Thanks for taking care of this!

src/main/java/io/jenkins/update_center/HPI.java
zbynek: SInce `apply` on Guava's `Function` is considered nullable and `PolicyFactory` does not override that annotation, SpotBugs sees this as a potential NPE source. The two ways to fix it seem to be 1) extracting to a method *without return type annotation* or 2) changing the declaring type to `java.util.function.Function`. I changed the solution from 1) to 2) and added comment.

PR #525, opened 2021-05-25
Formally deprecating plugins left behind in recent core PRs

basil:
In recent core PRs jenkinsci/jenkins#5320, jenkinsci/jenkins#5338, jenkinsci/jenkins#5521, and jenkinsci/jenkins#5526, we are intentionally leaving behind 15 plugins. These plugins are unmaintained, and they have not seen commits or releases in years. Even when we opened PRs against them, these PRs went unacknowledged. We have judged that the benefit of supporting these plugins is not worth the cost to ongoing maintenance.

We do our users a disservice by not explicitly marking these plugins as deprecated. These plugins are, de facto, unsupported. It may not be ideal, but this is the status quo and the reality. Users should be warned when installing these plugins for the first time, and users who have them already installed should be warned that these plugins may cause problems.

I am linking to a response from a Jenkins board member on an email thread I started about this topic. This is the most concise and official statement of our policy that I can find. If an improved statement materializes on jenkins.io, we can always update the link in the future. But for the sake of our users, we should deprecate these plugins now.

### DIFF
--- a/resources/deprecations.properties
+++ b/resources/deprecations.properties
@@ -159,7 +159,6 @@ poll-mailbox-trigger = https://git.io/JfaQM
 pretest-commit = https://wiki.jenkins.io/x/5gB-B
 # https://wiki.jenkins.io/display/JENKINS/pucm+plugin
 PUCM = https://wiki.jenkins.io/x/fIBVAw
-rad-builder = https://groups.google.com/g/jenkinsci-dev/c/9M2B6ZTKjI0/m/BChinuUVBAAJ
 # https://github.com/jenkins-infra/update-center2/pull/13
 rtc = https://git.io/JfaQ1
 # https://wiki.jenkins.io/display/JENKINS/Retry+Failed+Builds+Plugin
--- a/resources/deprecations.properties
+++ b/resources/deprecations.properties
@@ -18,7 +18,7 @@ azure-iot-edge = https://git.io/JtOjd
 blackduck-hub = https://git.io/JfaQa
 # https://wiki.jenkins.io/display/JENKINS/Black+Duck+Vulnerability+Installer+Plugin
 blackduck-installer = https://wiki.jenkins.io/x/nYHHB
-BlameSubversion = https://groups.google.com/g/jenkinsci-dev/c/9M2B6ZTKjI0/m/BChinuUVBAAJ
+BlameSubversion = https://github.com/jenkinsci/jenkins/pull/5320
 # https://github.com/jenkins-infra/update-center2/pull/521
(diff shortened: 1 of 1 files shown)


timja (approved):

olamy (approved):

resources/deprecations.properties
```diff
@@ -150,6 +159,7 @@ poll-mailbox-trigger = https://git.io/JfaQM
 pretest-commit = https://wiki.jenkins.io/x/5gB-B
 # https://wiki.jenkins.io/display/JENKINS/pucm+plugin
 PUCM = https://wiki.jenkins.io/x/fIBVAw
+rad-builder = https://groups.google.com/g/jenkinsci-dev/c/9M2B6ZTKjI0/m/BChinuUVBAAJ
```
daniel-beck, line 162:
```suggestion
```

The plugin still works (I assume -- at least the core PR didn't change anything about that). It can be installed from update sites and it will pull all needed dependencies. This only affects very narrow use cases around out-of-process plugin management.

> Screenshot 2021-05-26 at 10 27 35

daniel-beck (commented):
Re "deprecated": It's a bit annoying that we need to use the "deprecated" label for something that can trivially be un-deprecated again if someone shows up and resumes maintenance. OTOH, https://github.com/jenkinsci/jenkins/blob/677fbb2a81fa8e9a0bc37ad5eca56565612e1f2e/core/src/main/resources/hudson/PluginManager/table.properties#L63 is vague enough that a compatibility breaking issue qualifies. So I'm ±0 on using the label 😄 

Re references, I don't think an unresolved discussion on the dev list is a useful reference, even if a board member voices his personal opinion there. The respective core PR would make a better reference, I think (perhaps after limiting discussions to contributors); alternatively the changelog entry if it mentions the breaking change and lists affected plugins; perhaps can be changed to reference the LTS upgrade guide later once that exists.

jglick (approved):

basil:
> The respective core PR would make a better reference, I think (perhaps after limiting discussions to contributors); alternatively the changelog entry if it mentions the breaking change and lists affected plugins; perhaps can be changed to reference the LTS upgrade guide later once that exists.

Sounds good. I updated the link to refer to the relevant core PR for each plugin. If some better content eventually appears on jenkins.io in the form of the LTS upgrade guide or (better yet) an official statement regarding the project's plugin end-of-life (EOL) policy, I can go back and update the link.

batmat (approved):
thanks a lot Basil!

daniel-beck (approved):

PR #529, opened 2021-06-23
GeneXus plugin is not deprecated

jlamasrios:
Although included in the primary list of candidates for deprecation due to jenkinsci/jenkins#5320, the plugin was fixed on its 1.10 release on June 3, and removed from the list (see https://github.com/bitwiseman/jep/tree/00c184f97f40c5f79f8f83074c78dd4fb1e1ac43/jep/231)

The plugin does have mantainers and is not deprecated.


batmat (approved):
👍

PR #534, opened 2021-07-09
[INFRA-3023] A dd tuleap-api-2.3.0 to ignore list

olblak:


MarkEWaite (approved):
That matches with other techniques used in this file.

resources/artifact-ignores.properties
```diff
@@ -558,3 +558,6 @@ github-api-1.222
 
 # Messed up file, probably from manual upload
 ca-mat-performance-benchmarking-by-broadcom-1.2
+
+# INFRA-3023
+tuleap-api-2.3.0
```
jglick, line 563: If this does not work, perhaps try a more aggressive exclusion

```
tuleap-api
tuleap-oauth
tuleap-git-branch-source
```

PR #538, opened 2021-07-30
working-hours plugin has no content on plugin site

halkeye:


MarkEWaite (approved):

daniel-beck:
I'd like to see something on https://github.com/jenkinsci/working-hours-plugin/pulls or at least https://github.com/jenkinsci/working-hours-plugin/issues first before we add an override here; if the plugin is even minimally maintained, we shouldn't need to use overrides.

zbynek:
https://github.com/jenkinsci/working-hours-plugin/pull/41 was merged already, but not released for 2 years

daniel-beck:
@zbynek Thanks for that information.

PR #539, opened 2021-08-02
Remove duplication between deprecations and ignores

zbynek:
@daniel-beck this should make it easier to inform users about suspended plugins (no duplication between ignores and deprecations). General thoughts 👍 / 👎 ? If it's OK I'll double-check that no information got lost here.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -113,10 +113,8 @@ There are various reasons to need to do this, such as:
 * A plugin integrates with a service that has been shut down.
 
 Both use cases (entire plugins, or specific versions) are controlled via the file `resources/artifact-ignores.properties`.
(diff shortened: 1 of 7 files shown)


daniel-beck:
Oh that's a nice approach. I'll give it a closer look when I have some time, but from a quick look it looks very promising!

Also TODO (perhaps not now but before merge): Update README and resource file header comments.

daniel-beck:
Sorry for the delay here, my spare time is fairly limited right now.

daniel-beck (approved):
This is great, thank you very much!

That data migration 🤩

Will merge and release once you confirm you're OK with my changes.

zbynek:
@daniel-beck thanks for reviewing, your changes look good!

PR #545, opened 2021-09-12
All the plugins migrated to the new dump

halkeye:
I tried to keep the comments from the old redirects.

I couuld do the change via code, but i felt this was the right file to do it. Just looped through everything in that repo, which was generated by looping through the url documenation json file.


zbynek:
Wouldn't it be easier to change the code instead of the properties file? As in "if URL matches wiki.(jenkins.io|jenkins-ci.org) return dump-repo/artifact-id/blob/master/readme.txt"? There is already some  code for replacing jenkins-ci.org with jenkins.io...

halkeye:
easier? its already done. But yea, easier to maintain and stuff i donno. I guess code would mean when the repo gets updated, then the propery file doesn't need to be updated as well. 

I'm good either way I'll defer to @daniel-beck as the owner?

daniel-beck:
The problem with the override file is that it unconditionally overrides the location, making later plugin updates seemingly ineffective or even pointless. The ideal long-term state IMO is to not have (or need) overrides.

Having code rewrite a (now known to be bad) wiki URL to a (hopefully only temporarily needed) wiki dump URL seems preferable, as future updates of plugin metadata would be effective immediately, and not require updates in this repo.

halkeye:
I don't see this problem going away. These repos are many many years stale.

I'll make a new pr with a code changes

PR #549, opened 2021-09-27
Attempt to fix the regexp for nested view.

jtnord:
all versions prior to 1.21 with the exception of 1.19.1 are vulnerable.

@daniel-beck


resources/warnings.json
```diff
@@ -10362,7 +10362,7 @@
     "versions": [
       {
         "lastVersion": "1.20",
-        "pattern": "(1[.]1[0123479]|1[.]20|1[.][6-9])(|[.-].+)"
+        "pattern": "(1[.][0-9]|1[.]1[0-8]|1.19[^0-9.]|1[.]20)(|[.-].+)"
```
jtnord, line 10365: this is obviously wrong :-(
and reports 1.151 as broken.

jtnord:
tested at https://regex101.com/ with various numbers and wrapping the regexp with `^` and `$`

resources/warnings.json
```diff
@@ -10362,7 +10362,7 @@
     "versions": [
       {
         "lastVersion": "1.20",
-        "pattern": "(1[.]1[0123479]|1[.]20|1[.][6-9])(|[.-].+)"
+        "pattern": "(1[.][0-9]|1[.]1[0-8]|1[.]19(?![.])|1[.]20)(|[.-].+)"
```
daniel-beck, line 10365:
```suggestion
        "pattern": "(1[.]1[012347]|1[.]20|1[.][6-9])(|[.-].+)|(1[.]19)"
```

This would be the generated pattern we should have been using.

jtnord: I assume 1.2 and the other ower 1
X versions (pre 1
.6) where never distributed then?

daniel-beck: > never distributed

According to http://updates.jenkins.io/download/plugins/nested-view/ they do not exist.

(There's a limitation of the tool -- it only looks at currently distributed versions, so if a release got suspended, that would be missing. Not the case here.)

PR #552, opened 2021-11-09
Removing squash4jenkins due to security issue

mlafaix:
Security issue is JIRA SECURITY-2525

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -206,7 +206,7 @@ sms-notification                 # renamed to sms
 sonargraph-plugin = https://git.io/JJ5h1
 # deprecated (see Sonatype CLM: https://wiki.jenkins-ci.org/x/iYDPAw)
 sonatype-ci = https://wiki.jenkins.io/x/iYDPAw
-squash4jenkins                   # removal due to security issue JIRA SECURITY-2525
+squashtm-publisher                   # removal due to security issue JIRA SECURITY-2525
 ssh2easy-plugin                  # removal requested by plugin author, accidental rename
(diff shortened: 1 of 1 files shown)


resources/artifact-ignores.properties
```diff
@@ -206,6 +206,7 @@ sms-notification                 # renamed to sms
 sonargraph-plugin = https://git.io/JJ5h1
 # deprecated (see Sonatype CLM: https://wiki.jenkins-ci.org/x/iYDPAw)
 sonatype-ci = https://wiki.jenkins.io/x/iYDPAw
+squash4jenkins                   # removal due to security issue JIRA SECURITY-2525
```
Wadeck, line 209:
```suggestion
squashtm-publisher                   # removal due to security issue JIRA SECURITY-2525
```

From https://github.com/jenkinsci/squashtm-publisher-plugin/blob/5641bf160f66be9a93bca2f7f1555c3ae3de6d08/build.gradle#L22

Wadeck (commented):
We usually do not provide reference to non-yet-disclosed vulnerability to justify a plugin addition to the blocklist, but as you're also part of the maintainers of this plugin, it's fine, so 👍 from me
(I took https://henix-blog.medium.com/squash-autom-squash-devops-une-g%C3%A9n%C3%A8se-82a776208641 as a "proof" of ownership)

(not approving due to the wrong artifact id IIUC)

daniel-beck:
@mlafaix You really should have responded in Jira first. This is pretty bad form.

daniel-beck:
On-holding until we can properly publish an advisory.

daniel-beck:
https://www.jenkins.io/security/advisory/2021-11-12/#SECURITY-2525

PR #553, opened 2021-11-29
Ignore opentelemetry-1.0.0 that contains multiple versions for the same dependency

v1v:
### What

Ignore opentelemetry-1.0.0 that contains multiple versions for the same dependency.

### Issues

Caused by https://github.com/jenkinsci/opentelemetry-plugin/issues/230


MarkEWaite (approved):

daniel-beck:
Seems unnecessary since the release has been superseded?

PR #554, opened 2021-12-03
Discard hpi/jpi/war artifacts without corresponding pom file

daniel-beck:
https://updates.jenkins.io didn't update for several hours today (1:42 PM to 9:30 PM), discovered via https://groups.google.com/g/jenkinsci-dev/c/qFIOBwWCLyM/m/63OkleJfAQAJ

Root cause was that the tool cached a failure to download the `.pom` file of today's release of https://plugins.jenkins.io/git-automerger/

Since that is a stupid reason to stop working, this change discards all plugin `.hpi`/`.jpi` (and core `.war`) files without a corresponding `.pom` file. This typically should only filter out files for one run due to upload timing issues, or remove plugins that are improperly uploaded.

Untested.

### DIFF
--- a/src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
+++ b/src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
@@ -3,6 +3,8 @@ package io.jenkins.update_center;
 import com.alibaba.fastjson.JSON;
 import io.jenkins.update_center.util.Environment;
 import io.jenkins.update_center.util.HttpHelper;
+import java.util.Iterator;
+import java.util.function.Predicate;
 import okhttp3.Credentials;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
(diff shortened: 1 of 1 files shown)


src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
```diff
@@ -157,11 +160,26 @@ private void initialize() throws IOException {
...
     }
 
+    // We cannot use Collection#removeIf because we want to log this
+    private static <E> void removeIf(Collection<E> collection, Predicate<? super E> filter) {
+        final Iterator<E> each = collection.iterator();
+        while (each.hasNext()) {
+            final E current = each.next();
+            if (filter.test(current)) {
+                each.remove();
+                LOGGER.log(Level.INFO, "Removing artifact file without corresponding pom file: " + current);
```
oleg-nenashev, line 178:
```suggestion
                LOGGER.log(Level.WARNING, "Removing artifact file without corresponding pom file: " + current);
```

oleg-nenashev (approved):
I think higher logging level is justified in this case. It might be related to manual uploads and hence security concerns. CC @Wadeck

src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
daniel-beck: > It might be related to manual uploads and hence security concerns

Anyone entitled to release via m-r-p can choose not to and instead upload manually. No security concern really. (Artifactory offers to generate a POM anyway.)

PR #556, opened 2021-12-20
ignore monkit-plugin

halkeye:
see https://groups.google.com/g/jenkinsci-dev/c/1bIdZr8P_Ms


daniel-beck:
CC @MadsNielsen who's the sole maintainer.

oleg-nenashev (approved):
CC @ewelinawilkosz who might now the background. IIUC Eficode does not longer maintain Jenkins plugins. At least this is the response I got one year ago

PR #558, opened 2021-12-30
Mark sonar-gerrit warning as fixed

reda-alaoui:
The vulnerability was fixed in version 2.4.5


oleg-nenashev (approved):
FTR https://github.com/jenkinsci/sonar-gerrit-plugin/blob/master/CHANGELOG-TO-V2.md#245 . Note that the plugin now uses the CD versioning scheme. Should be okay after the merge

resources/warnings.json
```diff
@@ -6251,8 +6251,8 @@
     "url": "https://jenkins.io/security/advisory/2019-10-23/#SECURITY-1003",
     "versions": [
       {
-        "lastVersion": "2.3",
-        "pattern": ".*"
+        "lastVersion": "2.4.4",
+        "pattern": "(1[.]|2[.][0-3][.]|2[.]4[.][0-4]).*"
```
daniel-beck, line 6255: Excludes `2.4`. No need to mess with this yourself, @Wadeck can fix up the regex, we have a tool to do that.

reda-alaoui: Sorry but I don't understand what action I should take

reda-alaoui: Ok :)

reda-alaoui:
Ping :)

resources/warnings.json
Wadeck: Looking into this 👀. I think it was corrected in [353.v20e9cff705d1](https://github.com/jenkinsci/sonar-gerrit-plugin/releases/tag/353.v20e9cff705d1)

Wadeck: > Excludes 2.4

There is no such version in https://repo.jenkins-ci.org/ui/repos/tree/General/releases/org/jenkins-ci/plugins ;)

reda-alaoui: No, I can tell you it was fixed in 2.4.5 :)

reda-alaoui: https://github.com/jenkinsci/sonar-gerrit-plugin/commit/d86de84131660051de7a90195478761b7d087630

Wadeck (approved):
🚀 
Thanks for the correction @reda-alaoui

resources/warnings.json
Wadeck: You're too reactive, I saw my mistake in the meantime :D

daniel-beck: > There is no such version

Good point, I checked Git tags only.

daniel-beck: > No, I can tell you it was fixed in 2.4.5 :)

Right, my point is that 2.4 isn't matching the regex of versions with the vulnerability.

Wadeck: the 2.4.5 point was for me, because I first thought it was done a bit later with the credentials usage, but actually it was fixed in 2.4.5 with the Secret usage as first approach ^^

PR #559, opened 2022-01-01
Remove unused AWS API plugins

zbynek:
Since https://github.com/jenkinsci/aws-java-sdk-plugin/pull/540 these plugins are no longer released separately, they are all part of https://plugins.jenkins.io/aws-java-sdk-minimal . I think it's safe to just remove them since nothing depends on them, but if the preference is to deprecate for a while, remove later, I can do that instead.

@vlatombe OK?
CC @daniel-beck


daniel-beck:
Ideally a maintainer would sign off on the suspension. @vlatombe perhaps?

Vlatombe (approved):
LGTM

PR #561, opened 2022-01-11
--id parameter from CLI is ignored

nfalco79:
Actually the --id option that Main.java get from CLI is ignored.
There is also a TODO in UpdateCenterRoot bean file to use the id given in command line.

This PR fix the id in the update center produced. I know that is not intended to be used as private repository the PR does not do any different than fix something existing.

The same for connectionCheckUrl that is useful when www.google.com is not allowed in some countries or company proxy.


oleg-nenashev (changes requested):
Should be HTTPs for connectionCheckUrl  by default

nfalco79:
I agree, I just moved existing URL string. Let me fix in 5 minutes to https

nfalco79:
done

nfalco79:
I forgot a file.

daniel-beck:
@nfalco79 Sorry about the delay here! Does the last commit look OK to you?

nfalco79:
Do you mean if after rebase master this fix will works as expected? I do not know, I forked after a two months to generate a custom update center based on maven central as artifact repository.
I should test it again.

daniel-beck:
> Do you mean if after rebase master this fix will works as expected?

I don't know what you're trying to do; but this tool has used the Artifactory API for a long time. But I would expect `--id` and `--connection-check-url` to have the intended effect, yes.

nfalco79:
Initially I started to generate a custom update site using our artifactory (company repository). After that I changed company, so this PR remain just to fix the id parameter.

daniel-beck:
Well, crap, now I remember why I held off merging this. If only it were so easy…

PR #565, opened 2022-01-24
Suspend distribution of non-Java plugins

basil:
And it came to pass [at the end of thirty days](https://www.jenkins.io/blog/2021/12/22/deprecated-ruby-runtime/) that no users complained about the deprecation of non-Java plugins. And @basil said, may these plugins be suspended from distribution. And it was so, and no longer did the non-Java plugins roam the face of the earth.

CC @daniel-beck @halkeye


halkeye (approved):
lgtm

MarkEWaite (approved):

PR #568, opened 2022-01-25
Use maven-8 docker image

MarkEWaite:
## Use maven-8 docker image

Rely on the maven tool already being installed.  Also relies on Java 8 already being installed.

As suggested by @dduportal 

Fixes https://github.com/jenkins-infra/helpdesk/issues/2747

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,20 +5,14 @@ properties([
         buildDiscarder(logRotator(numToKeepStr: '10', artifactNumToKeepStr: '2')),
 ])
 
-node('linux') {
+node('maven-8') {
     stage('Prepare') {
         deleteDir()
         checkout scm
     }
 
     stage('Generate') {
-        withEnv([
-                "PATH+MVN=${tool 'mvn'}/bin",
-                "JAVA_HOME=${tool 'jdk8'}",
-                "PATH+JAVA=${tool 'jdk8'}/bin"
-        ]) {
-            sh 'mvn -e clean verify'
-        }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -5,23 +5,17 @@ properties([
         buildDiscarder(logRotator(numToKeepStr: '10', artifactNumToKeepStr: '2')),
 ])
 
-node('linux') {
+node('maven-8') {
```
dduportal, line 8: For now, trusted.ci only have a `maven-11` label: https://github.com/jenkins-infra/jenkins-infra/blob/production/hieradata/clients/trusted-ci.yaml#L49 (because JDK 11 is the default used so the default JAVA_HOME points toward JDK8).

I propose that we keep the current pipeline and close this PR as the instabilities of update_center2 jobs, on both ci.jenkins.io and trusted.jenkins.io were due to JFrog's incident.

@MarkEWaite can you confirm that my message makes sense and that I did not forget anything?

daniel-beck: Irrelevant because the job on trusted.ci does not use this Jenkinsfile.

dduportal: oh, what is the pipeline used on trusted ? (but good to know \o/)

daniel-beck: Freestyle job.

dduportal: <insert screaming emoji here>

MarkEWaite: Closing the pull request because the issues were unrelated to this change.  Thanks @dduportal

PR #570, opened 2022-01-31
Add deprecation notice for environment-dashabord

willwh:
As per title. This plugin is going to be retired.


MarkEWaite (approved):

zbynek:
It might be better to add the plugin to https://github.com/jenkins-infra/update-center2/blob/master/resources/deprecations.properties together with a URL of a deprecation notice (e.g. you can extend description of this PR with a simple explanation why is the plugin deprecated and what can current users do, then use URL of this PR as the deprecation notice).

PR #571, opened 2022-02-21
Suspend distribution of StarTeam plugin, it's closed source since 2015

daniel-beck:
Per their own documentation on the wiki page https://github.com/jenkins-infra/docker-confluence-data/blob/main/content/JENKINS/StarTeam.html currently rendered on https://plugins.jenkins.io/starteam/ :

> Screenshot 2022-02-21 at 14 21 08

The last release in https://github.com/jenkinsci/starteam-plugin/commits/master was 0.6.12 in 2013, so suspend distribution.

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -711,4 +711,4 @@ team-views = https://www.jenkins.io/security/plugins/#suspensions
 sinatra-chef-builder = https://www.jenkins.io/security/plugins/#suspensions
 
 # The plugin has been closed source since 2015
-starteam = https://wiki.jenkins.io/display/JENKINS/StarTeam
+starteam = https://github.com/jenkins-infra/update-center2/pull/571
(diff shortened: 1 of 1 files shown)


daniel-beck:
>  In case of questions => https://www.jenkins.io/doc/developer/publishing/preparation/#license

Well, the super old releases were MIT licensed, they just stopped doing those.

resources/artifact-ignores.properties
```diff
@@ -709,3 +709,6 @@ yammer = https://www.jenkins.io/blog/2021/12/22/deprecated-ruby-runtime/
 # https://www.jenkins.io/security/advisory/2022-02-15/
 team-views = https://www.jenkins.io/security/plugins/#suspensions
 sinatra-chef-builder = https://www.jenkins.io/security/plugins/#suspensions
+
+# The plugin has been closed source since 2015
+starteam = https://github.com/jenkins-infra/update-center2/pull/571
```
lemeurherve, line 714:
```suggestion
starteam = https://wiki.jenkins.io/JENKINS/StarTeam
```

lemeurherve: Without the "display" in the wiki URL there is no redirection to plugins.jenkins.io

MarkEWaite: > Without the "display" in the wiki URL there is no redirection to plugins.jenkins.io

I think this pull request is the better justification for suspending distribution, rather than pointing to the wiki page.  I find this pull request more easily read than the wiki page.

MarkEWaite (approved):

resources/artifact-ignores.properties
lemeurherve: I've suggested this as pointing to this plugin wiki page without being redirected to plugins.jenkins.io was a request from @daniel-beck on IRC.

PR #574, opened 2022-03-04
deprecated incapptic-connect-uploader-plugin

spherefoundry:
This plugin is no longer maintained by the incapptic Connect Team.


daniel-beck:
@spherefoundry Could you provide a public URL as reference of the deprecation, like an announcement? That way, plugins.jenkins.io will link to the URL as additional information, and current users of the plugin will be informed directly inside Jenkins as well.

spherefoundry:
Hey @daniel-beck I have added the URL to point to this PR and updated the description. I hope this will do?

daniel-beck:
Not great, but it'll do if there's nothing else 😉 

---

FTR I am in out of band contact with the maintainers, so even though the GH account looks unrelated, this request is legitimate.

PR #577, opened 2022-03-05
Revert "Suspend nexus-jenkins-plugin"

daniel-beck:
Reverts jenkins-infra/update-center2#576

Given https://groups.google.com/g/jenkinsci-dev/c/TZakKUjUm9Q/m/Pirq0fqEBAAJ it's possible this failed release is not (only) caused by the maintainer/release process side of things, so reverting this and instead only suspend the known broken release.


dduportal (approved):
👍

PR #578, opened 2022-03-21
Update artifact-ignores.properties - Remove eggplant-plugin 

pohkheng2:
This eggplant-plugin had been deprecated. We would like to request to remove the plugin from Jenkins plugin Update Center


daniel-beck:
@pohkheng2 Thanks for your submission. The URL here is intended to be one that provides information why the plugin is obsolete. Do you have documentation for that, perhaps?

pohkheng2:
@daniel-beck Thanks for quick response. We do not really have documentation for this eggplant-plugin removal. This plugin basically is too old and no longer be able to work with any Eggplant software in the market.

Do let us know if you need any additional information.

ryantestplant (approved):

PR #582, opened 2022-03-29
Supersede ui-samples-plugin

NotMyFault:
FYI @timja


daniel-beck:
Should get a URL, e.g. to a public discussion, issue, or other reference about the artifact ID change. Then Jenkins will show this to current users.

NotMyFault:
Tim wants to propose a post on the mailing list I'll link once it's live.

timja (approved):

daniel-beck:
That reference does not actually explain anything about the artifact ID change, it just announces publication of a new thing. I guess this can still be merged given the nature of the plugin and low popularity, but still seems like a miss.

PR #585, opened 2022-04-13
Formally deprecating some plugins left behind by JEP-233

basil:
These plugins were left behind in JEP-233 and no longer work on the latest LTS. I have not heard any complaints about any of them, and most of them haven't been updated in years. Many have existing unresolved security vulnerabilities.

The code comments say "Shorten URLs if and only if the URLs are to github.com (git.io)", but I haven't done this because git.io says "Sorry, we're no longer accepting new links to shorten."


daniel-beck:
Thanks! What's the source for this list? I would have guessed https://issues.jenkins.io/browse/JENKINS-65988 but there are more open plugin issues than just these added here.

basil:
> What's the source for this list? I would have guessed https://issues.jenkins.io/browse/JENKINS-65988

You guessed correct!

daniel-beck:
To clarify, we're deliberately only deprecating a subset of the plugins not compatible with current Jenkins? What distinguishes those plugins added here from others? E.g. why not also deprecate `aws-beanstalk-publisher-plugin`?

basil:
BTW this kind of attention to detail would be of great assistance in [JENKINS-68251](https://issues.jenkins.io/browse/JENKINS-68251).

Three different users have complained about AWS Elastic Beanstalk Publisher, so I am going to adopt and release it as in jenkins-infra/repository-permissions-updater#2487.

SCM Sync Configuration is already deprecated.

SmileHub Notifier was released less than a year ago, so I am still holding out hope that the maintainer will reappear. I could go either way on this one, though.

I don't really feel that I understand what is going on with "Google Authenticated Source" to be able to confidently deprecate it. Sounds like it could actually be useful, it was released only 2 years ago, and it has almost 1,000 installations (far more than all these others), so I am also holding out hope that a maintainer will reappear. I could go either way on this one, though.

daniel-beck:
Thanks for your response, that makes sense.

----

https://www.jenkins.io/blog/2021/11/09/guava-upgrade/ or even https://issues.jenkins.io/browse/JENKINS-65988 look like the better reference to add here for admins. Going to https://github.com/jenkinsci/jenkins/pull/5707 what jumps out immediately to me is the proposed changelog, which states (bold highlight mine)

> Jenkins has upgraded the [Guava](https://guava.dev/) library from [11.0.1](https://github.com/google/guava/releases/tag/v11.0.1) (released on January 9, 2012) to [31.0.1](https://github.com/google/guava/releases/tag/v31.0.1) (released on September 27, 2021). **Plugins have already been prepared to support the new version of Guava**; use the Plugin Manager to upgrade all plugins before and after upgrading Jenkins.

leaving me to wonder why this causes deprecation.

Thoughts?

basil:
Done in commit 06db9cd.

daniel-beck (approved):
Thanks!

PR #586, opened 2022-04-16
Deprecate AdoptOpenJDK plugin

NotMyFault:
AdoptOpenJDK as service itself is deprecated since late 2020 in favor of the transition to the Eclipse foundation.

### DIFF
--- a/resources/deprecations.properties
+++ b/resources/deprecations.properties
@@ -37,3 +37,4 @@ vertx = https://git.io/Jn6WI
 vs-code-metrics = https://git.io/Jn6Co
 # https://github.com/jenkinsci/jenkins/pull/5320
 vss = https://git.io/Jn6Co
+adoptopenjdk = https://blog.adoptopenjdk.net/2021/03/transition-to-eclipse-an-update/


zbynek:
CC @hrmohr as the plugin maintainer

MarkEWaite (changes requested):
I think that we should update the plugin to support Eclipse Temurin rather than deprecating the plugin.  The plugin ID would remain the same but the downloads would come from https://adoptium.net/temurin/releases/

We use Eclipse Temurin for infrastructure on ci.jenkins.io, infra.ci.jenkins.io, trusted.ci.jenkins.io, and weekly.ci.jenkins.io.

We don't use this plugin to download the JDK on that infrastructure, but this plugin is a reasonable alternative to the ["Oracle Java SE Development Kit Installer" installer](https://plugins.jenkins.io/jdk-tool/) and avoids the licensing issues associated with the JDK downloads from Oracle

NotMyFault:
While this plugin may offers an alternative to the Oracle installer plugin, rebuilding it for the Adoptium website reintroduces the same error prone setup depending on the website layout like in the current state.

There are countless (beginner-friendly) guides for and from major vendors like Adoptium, Liberica, MS or Zulu for a plenty of OS's and package managers to choose from. It's unlikely that this plugin continues to serve a demand in the future, considering the direction this ecosystem evolves into, for sake of having a hacky implementation in the Jenkins interface.

daniel-beck (dismissed):
Please merge with master after recent cleanup to this file, and add the line at the appropriate sorted location (while it's ambiguous, I'd say it should be sorted case-insensitively).

basil:
I concur with Mark. Whether the original code was fragile or not, it served a purpose (making it effortless for users to use a free JDK with Jenkins). Taking on the burden of dealing with the fragility makes life harder for maintainers but easier for users. I believe we should always put users first. My vote would be to retain the plugin and adapt it to Eclipse Temurin. Of course, if nobody steps up to do this, then deprecating it would be a (sad, for our users) alternative.

daniel-beck:
> I think that we should update the plugin to support Eclipse Temurin

It is unclear to me what would need changing other than the plugin display name.

The downloads for the plugin are Temurin (from e.g.  https://github.com/adoptium/temurin11-binaries), and seem to be recent releases too, so what exactly is the problem with the plugin?

basil:
I have not tested the plugin. At least the display name needs to be changed, other changes (if any) to be determined after testing.

daniel-beck:
I don't think there's enough justification here to deprecate, if the plugin works and does what its users would expect it to do, and all that's needed is changing the display name.

basil:
I changed the display name in https://github.com/jenkinsci/adoptopenjdk-plugin/pull/23 and migrated the crawler to the new API in https://github.com/jenkins-infra/crawler/pull/124.

PR #589, opened 2022-04-26
Add adopt-this-plugin label to svnmerge

hugueschabot:
As the current maintainer of the `svnmerge` plugin, I would like to mark it for adoption.


daniel-beck:
I added the label to the GitHub repo, which is the preferable approach (better self service).

alecharp (approved):

PR #591, opened 2022-05-07
Revert JENKINS-20679

basil:
Reverts #253. See [the mailing list thread](https://groups.google.com/g/jenkinsci-dev/c/V2CH4pNQitQ/m/ZNmNb0x7BwAJ) along with https://github.com/jenkinsci/jenkins/pull/6549 and https://github.com/jenkinsci/maven-hpi-plugin/pull/341. This is dead code, so we intend to stop producing it from `maven-hpi-plugin` and to stop consuming it from `jenkins` and `update-center2`. To test that dropping `minimumJavaVersion` from the Update Center won't affect older versions of Jenkins, I simulated this condition by setting `this.minimumJavaVersion` to `null` in `jenkinsci/jenkins/core/src/main/java/hudson/model/UpdateSite.java`. I refreshed the update center metadata successfully and updated several plugins without any issues.


daniel-beck:
FYI this is currently live through https://github.com/jenkins-infra/update-center2/tree/use-3.12.2; will make it permanent with a real release tomorrow unless there are complaints. (I don't expect there will be.)

basil:
Thank you Daniel! I don't expect any problems either given my testing, but I appreciate the heads up and will try to keep an eye on things just in case.

alecharp (approved):

PR #531, opened 2021-07-01
deprecate performance-signature-dynatrace

rpionke:
The product Dynatrace AppMon is deprecated since 31.03.2021: (https://www.dynatrace.com/support/doc/appmon/)
I also added the successor of Dynatrace AppMon: https://plugins.jenkins.io/performance-signature-dynatracesaas/

FFR:
image

PR #593, opened 2022-05-18
Deprecate two Dynatrace plugins

zbynek:
Like https://github.com/jenkins-infra/update-center2/pull/531, but not using the dead git.io service


basil:
Thank you!

PR #594, opened 2022-05-19
[JEP-230] Proactively block `instance-identity` 3.1

jglick:
Want to be able to release https://github.com/jenkinsci/instance-identity-plugin/pull/22 without it appearing on the UC, at least for now.

(I am not sure I have permission to release by the way? Says `@core` but I am not sure if I am in that group or not. How would I check? Recently I have been using JEP-229 exclusively for releases, which I _could_ do here but it would be tricky since I need to predict the version number in advance.)

@timja

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -524,9 +524,10 @@ kubesphere-extension-0.2.0
 sshd-1.0
 sshd-1.1
 sshd-1.2
-# pending https://github.com/jenkinsci/jenkins/pull/5304
+# pending completion of JEP-230
 instance-identity-3.0
 instance-identity-3.0.1
+instance-identity-3.1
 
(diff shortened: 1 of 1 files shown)


daniel-beck:
> (I am not sure I have permission to release by the way? Says `@core` but I am not sure if I am in that group or not. How would I check?

Team definitions are in https://github.com/jenkins-infra/repository-permissions-updater/tree/master/teams

jglick:
https://github.com/jenkins-infra/repository-permissions-updater/blob/088c793cdaf26fe14c1ac6a6a2d2e2c82e81a59d/teams/core.yml#L9 OK thanks.

timja:
It's a list of all active core maintainers really, but any core maintainer can PR to be in there.

PR #596, opened 2022-05-29
Fix cas-plugin version pattern for SECURITY-2387

fcrespel:
The version pattern in warnings.json for SECURITY-2387 affecting cas-plugin was too broad, basically 1.x instead of any version up to 1.6.0. The vulnerability was fixed in 1.6.1, so this PR updates the pattern accordingly.

For reference: [cas-plugin changelog](https://github.com/jenkinsci/cas-plugin/blob/master/CHANGELOG.md#161---2021-06-29) and [security advisory](https://www.jenkins.io/security/advisory/2021-06-30/#SECURITY-2387).


daniel-beck:
Sorry about that! I checked what happened, and my best explanation is a (now) known quirk in our tooling. Unsure since when we know about it, this might be an early occurrence we missed.

PR #599, opened 2022-06-03
Block `workflow-api` at `1162.va_1e49062a_00e`

jglick:
Realized post release that https://github.com/jenkinsci/workflow-api-plugin/pull/221 would be offered to 2.289.x–2.319.x UCs without https://github.com/jenkinsci/workflow-cps-plugin/pull/534 which it requires to avoid some anomalous behavior. Best to block it from those lines, and let 2.332.x— pick up https://github.com/jenkinsci/workflow-api-plugin/pull/226. @dwnusbaum @car-roll


dwnusbaum (approved):

jglick:
Seems to have caused https://issues.jenkins.io/browse/JENKINS-68727.

PR #603, opened 2022-06-18
Deprecate Maven Repository Scheduled Cleanup

basil:
JNR was removed in https://github.com/jenkinsci/jenkins/pull/6323. https://github.com/jenkinsci/maven-repo-cleaner-plugin/pull/4 dropped the dependency on JNR in Maven Repository Scheduled Cleanup, and that PR was merged on June 26, 2018; however, it has not been released.


MarkEWaite (approved):
@olamy I think that it is the right choice to deprecate this plugin rather than attempt to update it to release a new version that removes the dependency on jnr-posix.

https://stats.jenkins.io/pluginversions/maven-repo-cleaner.html shows there are 3200 installations of the plugin spread widely across Jenkins versions.

I'm noting in the Jenkins 2.346.1 upgrade guide that a 2.346.1 compatible version of the plugin has not been released.  Does that work OK for you?

PR #605, opened 2022-06-24
Switch cache to SHA-256

jglick:
Fixes https://github.com/jenkins-infra/helpdesk/issues/1968 by making all cache files have 64-char names. **Untested.**

PR #609, opened 2022-06-27
Split cache into sha prefix based directories

daniel-beck:
Amends https://github.com/jenkins-infra/update-center2/pull/605.

Now that we use a hash for the cache, we can split it by prefix, thereby limiting the number of files in each directory.

Two hex digits makes for 256 top-level directories, with around 190 files in each based on the current cache size.

Proposed migration script:

```
/home/jenkins/agent-workspace/workspace/update_center/caches/artifactory
find . | cut -c3- | while read -r LINE ; do
  RAW="$( echo -n "$LINE" | base64 -d )"
  SHA="$( echo -n "$RAW" | sha256sum | cut -c1-64 )"
  SHA_PREFIX="$( echo "$SHA" | cut -c1-2 )"
  mkdir -p "$SHA_PREFIX"
  mv -v "$LINE" "$SHA_PREFIX/$SHA"
done
```

### DIFF
--- a/src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
+++ b/src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
@@ -228,11 +228,18 @@ public class ArtifactoryRepositoryImpl extends BaseMavenRepository {
     }
 
     private File getFile(final String url) throws IOException {
+        // TODO remove old base64 based cache paths once the cache is migrated
         String urlBase64 = Base64.encodeBase64String(new URL(url).getPath().getBytes(StandardCharsets.UTF_8));
         File cacheFile = new File(cacheDirectory, urlBase64);
         if (!cacheFile.exists()) {
             // Preferred new location (guaranteed maximum filename length):
-            cacheFile = new File(cacheDirectory, DigestUtils.sha256Hex(url));
+            final String sha256 = DigestUtils.sha256Hex(url);
+            final String sha256prefix = sha256.substring(0, 2); // to limit number of files in top-level directory
+            final File cachePrefixDir = new File(cacheDirectory, sha256prefix);
+            if (!cachePrefixDir.exists() && !cachePrefixDir.mkdirs()) {
+                LOGGER.log(Level.WARNING, "Failed to create cache prefix directory " + cachePrefixDir);
+            }
+            cacheFile = new File(cachePrefixDir, sha256);
         }
 
         if (!cacheFile.exists()) {


dduportal (approved):
Good idea, that would put less impact on the underlying file system (easier to run shell commands in the file tree, better page caching, less pressure on the I/Os).

Shellcheck agrees with the syntax of the proposed migration script.

Tested the migration script on a local copy, looks good to me (at first glance of the result).


+100 with the evaluation of size


Many thanks!

src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
```diff
@@ -228,11 +228,18 @@ private InputStream getFileContent(String url) throws IOException {
...
         if (!cacheFile.exists()) {
             // Preferred new location (guaranteed maximum filename length):
-            cacheFile = new File(cacheDirectory, DigestUtils.sha256Hex(url));
+            final String sha256 = DigestUtils.sha256Hex(url);
+            final String sha256prefix = sha256.substring(0, 2); // to limit number of files in top-level directory
+            final File cachePrefixDir = new File(cacheDirectory, sha256prefix);
+            if (!cachePrefixDir.exists() && !cachePrefixDir.mkdirs()) {
+                LOGGER.log(Level.WARNING, "Failed to create cache prefix directory " + cachePrefixDir);
+            }
+            cacheFile = new File(cachePrefixDir, sha256);
```
jglick, line 242: Note that this repeats the prefix. Not sure if that was intentional.

src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
```diff
@@ -228,11 +228,18 @@ private InputStream getFileContent(String url) throws IOException {
...
+        // TODO remove old base64 based cache paths once the cache is migrated
         String urlBase64 = Base64.encodeBase64String(new URL(url).getPath().getBytes(StandardCharsets.UTF_8));
         File cacheFile = new File(cacheDirectory, urlBase64);
         if (!cacheFile.exists()) {
             // Preferred new location (guaranteed maximum filename length):
-            cacheFile = new File(cacheDirectory, DigestUtils.sha256Hex(url));
+            final String sha256 = DigestUtils.sha256Hex(url);
+            final String sha256prefix = sha256.substring(0, 2); // to limit number of files in top-level directory
+            final File cachePrefixDir = new File(cacheDirectory, sha256prefix);
+            if (!cachePrefixDir.exists() && !cachePrefixDir.mkdirs()) {
```
jglick, line 239: BTW switch to NIO!

jglick (commented):
Seems like a good idea.

PR #610, opened 2022-06-27
Use 3.11, note 3.11 in the changelog

daniel-beck:


dduportal (approved):
\o/

CHANGELOG.adoc
```diff
@@ -2,6 +2,11 @@
 
 NOTE: This changelog only covers the actual tool (`update-center2-x.y.z.jar`), not changes to the wrapper scripts or metadata.
 
+== 3.11 (2022-06-27)
+
+* Switch Artifactory cache from base64 based file names to sha256 to support longer file paths (https://github.com/jenkins-infra/update-center2/pull/605[#605], https://github.com/jenkins-infra/update-center2/pull/609[#609]).
```
jglick, line 7: #609. (Why does this have a changelog at all?)

daniel-beck: I don't understand either part of your comment…?

jglick: First part: just mentioning the PR here so that there is a proper link & backlink.

Second part: just wondering why this repository would have a changelog and a formal version number when AFAIK it is for the exclusive use of the Jenkins infrastructure. Who would read the changelog? Just the one person making the change go live by some bump somewhere in some GitOps manifest?

daniel-beck: > when AFAIK it is for the exclusive use of the Jenkins infrastructure.

Two reasons (for me):

* Since we do have actual versions starting from 3.0 as a side effect of a build time optimization, it's just good practice to note the changes, even if just for our future selves.
* While it's not _supported_, there are other users of it.

> Just the one person making the change go live by some bump somewhere in some GitOps manifest?

That would be nice (maybe) but in reality, this (specifically the shell script change) _is_ the bump. 😁

dduportal: >  Just the one person making the change go live by some bump somewhere in some GitOps manifest

Changelog is useful or knowledge sharing (and for people like me who tend to forget thing after 3-4 months) 🤗 😅

PR #615, opened 2022-07-05
deprecate hpe-network-virtualization-plugin

itayka:


resources/label-definitions.properties
```diff
@@ -439,7 +439,7 @@ housekeeper=post-build
 hp-application-automation-tools-plugin=misc
 hp-operations-orchestration-automation-execution-plugin=misc external
 hp-quality-center=post-build
-hpe-network-virtualization=misc
+hpe-network-virtualization=deprecated
```
daniel-beck, line 442: This is unnecessary and unused if the plugin is suspended.

resources/artifact-ignores.properties
```diff
@@ -106,6 +106,7 @@ HiddenParameter                  # renamed to hidden-parameter
 hipchat-plugin                   # renamed to hipchat
 # service discontinued -- https://wiki.jenkins.io/display/JENKINS/HockeyApp+Plugin
 hockeyapp = https://wiki.jenkins.io/x/7oPPAw
+hpe-network-virtualization-plugin	# end of life
```
daniel-beck, line 109: Could you provide a reference for why this plugin is deprecated? If so, that URL should be the value to this key, as it will be provided to users of the plugin as an explanation.

Otherwise there will not be a deprecation notice at all, the plugin will just be gone.

resources/label-definitions.properties
itayka: Would you like me to roll back this change?

resources/artifact-ignores.properties
daniel-beck:
```suggestion
hpe-network-virtualization-plugin = https://github.com/jenkins-infra/update-center2/pull/615#discussion_r913923468
```

😁

itayka: I am not sure my managers will love this 😂. Everything a customer sees has to go through technical writer. I prefer to leave it as is.

daniel-beck: There is no solution we support in which the plugin would be suspended from distribution AND marked deprecated without pointing to a custom URL. The reason is that suspending distribution removes its page from plugins.jenkins.io, and there's no way to get anywhere without having a URL. WIth a URL, there'll be a deprecation notice both inside Jenkins and on plugins.jenkins.io, like https://plugins.jenkins.io/xtrigger/

I'd usually suggest a one line notice in the project's readme and to link there, but this plugin is still on the read-only wiki, and would need a new release to switch the documentation source to GitHub.

Does the related product have a deprecation/EOL notice we could hijack for the Jenkins plugin integrating with it?

itayka: I don't think so. I am not sure when exactly this product reached its EOL. But, the trigger for the plugin removal were many security issues discovered by the Jenkins security team. One of my managers was notified about it. Since I am the only contributor/owner of the plugin, I was asked to remove it.

itayka: I have a link about the EOL of our stand alone product. I will make the changes

itayka: I added a URL. Please approve and merge

resources/label-definitions.properties
daniel-beck: Ideally, given the removal of the plugin,
```suggestion
```

daniel-beck:
This should be effective inside Jenkins (with "Refresh Data") within a few minutes, and on plugins.jenkins.io in a few hours.

PR #616, opened 2022-07-12
[SECURITY-2521] Note fix in version 351

daniel-beck:
Resolves https://github.com/jenkins-infra/helpdesk/issues/3050

### DIFF
--- a/resources/warnings.json
+++ b/resources/warnings.json
@@ -11133,7 +11133,7 @@
     "versions": [
       {
         "lastVersion": "336.v182c0fbaaeb7",
-        "pattern": ".*"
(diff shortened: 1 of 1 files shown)


jetersen (approved):

jetersen:
Not sure how this regex is read but the regex is pretty open

https://regex101.com/r/NjD19Z/1

daniel-beck:
> Not sure how this regex is read but the regex is pretty open
> 
> https://regex101.com/r/NjD19Z/1

The entire thing needs to match, so wrap in `^(…)$` to get the comparable effect.

PR #620, opened 2022-07-26
Replace broken wiki short URLs with full URLs

daniel-beck:
Short URLs were nice to save bytes on `update-center.json`, but they no longer work.


resources/artifact-ignores.properties
```diff
@@ -219,29 +219,29 @@ sonargraph-plugin = https://github.com/jenkins-infra/update-center2/pull/405
 sonatype-ci = https://wiki.jenkins.io/x/iYDPAw
```
daniel-beck, line 219: https://wiki.jenkins.io/display/JENKINS/63930505 perhaps?

MarkEWaite: +1 from me to use that replacement.

MarkEWaite (approved):
Thanks for fixing the broken links!

dduportal (approved):

dduportal:
@daniel-beck May I let you merge if ready to go?

PR #569, opened 2022-01-26
Simplify symlink creation, update dependencies

MarkEWaite:
## Update dependencies

Use the current parent pom and other dependencies.  Update the spotbugs exclusions for new warnings reported with the most recent spotbugs.  Create symbolic links and hard links with Java NIO methods instead of making calls to the operating system.  The change to Java NIO calls avoids a spotbugs warning related to ProcessBuilder command injection.

Exclude okhttp and annotation-api from fastjson because newer versions are being provided by other dependencies.

Update owasp-java-html-sanitizer to 20211018.2 to resolve a security warning.

- Use fastjson 1.2.79, not 1.2.73
- Use okhttp-urlconnection 4.9.3, not 4.8.0
- Use commons-io 2.11.0, not 2.7
- Use commons-lang3 3.12.0, not 3.11
- Use commons-codec 1.15, not 1.14
- Use support-log-formatter 1.1, not 1.0
- Use version-number 1.9, not 1.7
- Use ant 1.10.12, not 1.10.11
- Use owasp-java-html-sanitizer 20211018.2, not 20200713.1
- Use spotbugs-annotations 4.5.3, not 4.1.1
- Use jsoup 1.14.3, not 1.14.2
- Use kotlin-stdlib-common 1.6.10, not 1.3.72
- Use mockito-core 4.3.1, not 3.4.6
- Use junit 4.13.2, not 4.13.1
- Use mockwebserver 4.9.3, not 4.8.0
- Use parent pom 1.70, not 1.54
- Use spotbugs 4.5.3.0, not 4.2.0
- Use surefire 3.0.0-M5, not 2.18.1
- Exclude okhttp and annotation-api from fastjson
- Exclude several spotbugs warnings
- Allow catch of NPE in getMetaData()
- Exclude command injection vulnerability
- Replace ProcessBuilder with Java 8 file calls
- Exclude path traversal spotbugs warnings

PR #621, opened 2022-07-27
Block mergebase-sca because its version numbers cause trouble

daniel-beck:
We invoke `ln` with the version number as second arg without `--`, so `ln` thinks this is a flag. That'll be the proper fix (or, even properer, apply https://github.com/jenkins-infra/update-center2/pull/569), but for now this needs to work again to unblock today's security advisory delivery.

FYI @delanAtMergebase, we'll readmit your plugin later today or tomorrow once we've made this tool more robust.

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -744,3 +744,6 @@ autocomplete-parameter = https://www.jenkins.io/security/plugins/#suspensions
 workflow-api-1162.va_1e49062a_00e
 # https://issues.jenkins.io/browse/JENKINS-68727
 workflow-durable-task-step-1144.vd77b_57189936
+
+# Block mergebase-sca because its version numbers cause trouble with this tool: https://repo.jenkins-ci.org/artifactory/releases/com/mergebase/mergebase-sca/
+mergebase-sca


lemeurherve (approved):

lemeurherve:
FTR, here is the associated error log:

image

delanAtMergebase:
Thanks for the heads up. I'll re-deploy a more reasonable version number - I believe I used the incrementalize plugin incorrectly leading to this very strange version.

delanAtMergebase:
I've deployed version 1-rc21.6d7cfea04562 which no longer starts with a '-' and hopefully mitigates this issue. Let me know if I need to submit a request to delete the offending version.

daniel-beck:
@delanAtMergebase Please note that the version is still wrong in that it claims to be an 'rc'. You mixed instructions for incrementals (basically deploying pull request builds), which use this format, and CD, which does not.

delanAtMergebase:
I just wanted to follow up on this - we deployed a normal version without 'rc'. Do we have to get the offending version removed before we can get re-admitted?

daniel-beck:
> I just wanted to follow up on this - we deployed a normal version without 'rc'. Do we have to get the offending version removed before we can get re-admitted?

Oops, sorry about the delay here. I'll restore distribution of your plugin within a day.

delanAtMergebase:
@daniel-beck Hi Daniel,

Are we able to add distribution to this plugin? If there's any remaining issues I'll resolve them ASAP. We'd like to start actively advertising it. Thank you!

PR #626, opened 2022-07-28
[JENKINS-69164] Suspend distribution of katalon plugin

daniel-beck:
[JENKINS-69164](https://issues.jenkins.io/browse/JENKINS-69164)

CC @MarkEWaite @halkeye as the more active board members.

CC @duyluonganh @HieuBui419 who seem to be the plugin maintainers.


halkeye (approved):
Good by me

Its pretty easy to undo so I don't think they need a lot of warning

daniel-beck:
It's been a week with no response, so I merged this.

PR #627, opened 2022-07-29
[SECURITY-2220] Fixed by plot 2.1.11

offa:
[Plot v2.1.11](https://github.com/jenkinsci/plot-plugin/releases/tag/plot-2.1.11) fixes the issue.

Just to get this right:

- `lastVersion` is the last *affected* version (<= 2.1.10)
- any version _matching_ the `pattern` is affected (everything >= 2.1.11 shouldn't match respectively)


daniel-beck:
@offa Thanks, I'll confirm the issue is fixed and once that's done, merge this PR. Both of your assumptions are correct.

resources/warnings.json
```diff
@@ -13035,7 +13035,7 @@
     "versions": [
       {
         "lastVersion": "2.1.10",
-        "pattern": ".*"
+        "pattern": "([01][.].*|2[.]0[.].*|2[.]1.[0-9]|2[.]1[.]10)"
```
daniel-beck, line 13038: FTR our (sadly not public) script generates, based on currently published releases and the usual format that allows any suffixes,
```suggestion
        "pattern": "(1|2[.]0|2[.]1[.][0-9]|2[.]1[.]10)(|[.-].+)"
```

There's a bug in your expression with the non-escaped `.` in `2[.]1.[0-9]`. Version 2.100 would be considered vulnerable.

offa: :man_facepalming: Thanks!

PR #628, opened 2022-08-02
[SECURITY-2053] Fixed by http_request plugin 1.16

MarkEWaite:
## [SECURITY-2053](https://www.jenkins.io/security/advisory/2022-07-27/#SECURITY-2053) Fixed by http_request plugin 1.16

[SECURITY-2053](https://www.jenkins.io/security/advisory/2022-07-27/#SECURITY-2053) is resolved by [release 1.16](https://github.com/jenkinsci/http-request-plugin/releases/tag/http_request-1.16) of the http_request plugin.


dduportal:
Hello! What is the status of this PR, as ci.jenkins.io seems to have this message.

MarkEWaite:
I believe that the changes in [release 1.16](https://github.com/jenkinsci/http-request-plugin/releases/tag/http_request-1.16) need to be reviewed by the security team to confirm that the issue is fully resolved.  Once the issue is confirmed fully resolved, then this can be merged (assuming I provided the correct regex).

resources/warnings.json
```diff
@@ -13282,7 +13282,7 @@
     "versions": [
       {
         "lastVersion": "1.15",
-        "pattern": ".*"
+        "pattern": "(0|1[.][0-9]|0|1[.][0-9][.][0-9]+|1[.]1[0-5])"
```
daniel-beck, line 13285: This regex includes the non-existing version `0` twice. It's not _wrong_, just unnecessary. Additionally, this expression is quite repetitive. Instead of using an optional part to cover both 1.x and 1.x.y, it has two separate branches for these.

The regex in the usual format of prefixes allowing any suffix not producing false positives would be the shorter:

```suggestion
        "pattern": "(1[.][1-9]|1[.]1[0-5])(|[.-].+)"
```

daniel-beck:
I confirm the issue is resolved in 1.16. I suggest you still clean up/simplify the regex, perhaps with the provided suggestion.

PR #629, opened 2022-08-04
[JENKINS-69229] Suspend distribution of trilead-api-1.71.v9e7860a_67a_df

daniel-beck:
While the Java 11 requirement is documented in https://github.com/jenkinsci/trilead-api-plugin/releases/tag/1.71.v9e7860a_67a_df, the plugin does not require a Java 11 only release of Jenkins, and does not set the incompatibility flag (whose message wouldn't be great but at least folks would read the changelog).

I suggest we do the following:

* Suspend this release of `trilead-api`
* @kuisathaverat bumps the core dependency of `trilead-api` to 2.357 so it's only offered to Java 11 only update sites.

Thoughts?

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -749,3 +749,6 @@ workflow-durable-task-step-1144.vd77b_57189936
 mergebase-sca
 
 katalon = https://issues.jenkins.io/browse/JENKINS-69164
+
+# Incompatible with Java 8, but does not require 2.357+ to restrict itself to Java 11 only releases of Jenkins
+trilead-api-1.71.v9e7860a_67a_df


jglick:
https://issues.jenkins.io/browse/JENKINS-69229?

daniel-beck:
> https://issues.jenkins.io/browse/JENKINS-69229?

Yes, had it in the branch name, now also in the title 👍

dduportal (approved):
Update center build are now building properly again (infra team just updated the agent configuration to avoid issues due to trilead itself)

daniel-beck:
I was hoping for feedback from @kuisathaverat before merging: If he disagrees with the proposals from me or James, this is useless, future releases will just have the same problem. Additionally, the bom is currently uninstallable, plugins using the latest bom will not be installable…

dduportal:
> I was hoping for feedback from @kuisathaverat before merging: If he disagrees with the proposals from me or James, this is useless, future releases will just have the same problem. Additionally, the bom is currently uninstallable, plugins using the latest bom will not be installable…

I'm sorry, I misunderstood the impact. My goal when merging was to make sure that other end-users were not bitten like we were if they used JDK8 agent (even it is a good indicator for them to update their agent configs 😅).

The BOM issue is not clear to me though (why is this PR able to impact the BOM like this? that is a legit and candid question).

I'm available to rollback if needed.

daniel-beck:
Reminder that I maintain this repo, if I wanted this urgently I would have merged myself 😉 Same with the rollback.

kuisathaverat:
It is fine to suppend this version from distribution, it does not have any improvement, only removes a library that makes a false positive on Security scanners. the next version will be only installable in the next LTS that will no support Java 8

PR #632, opened 2022-08-10
Fix symlink creation when it already exists

daniel-beck:
The Java API does not overwrite existing symlinks, so we need to delete what's there first.

Also add better logging when things go wrong.

Amends #569.


MarkEWaite (approved):

MarkEWaite:
Sorry that I missed that API change in the switch from `ln -s` to use NIO.

PR #635, opened 2022-08-11
Do not wipe options with default values when using --arguments-file

daniel-beck:
Amends #561.

Without this, using `--arguments-file` with reset the arguments to `null`.

There's redundancy between the annotation and the actual field. This isn't great but I couldn't think quickly of a way to record default values, and args4j prints field values as part of `--help` output, so leaving the fields at their empty default would be misleading.


nfalco79:
Now I understand what you mean.

daniel-beck:
Yeah, my comment to your PR was pretty useless. Sorry about that!

We generate dozens of version specific update sites in infra, so doing everything in a single tool invocation optimizes execution time.

PR #639, opened 2022-08-12
SECURITY-2413 in clif-performance-testing is fixed

dillense:
[version 70.v7214dfdb_32ec](https://github.com/jenkinsci/clif-performance-testing-plugin/releases/tag/70.v7214dfdb_32ec) fixes [the issue](https://www.jenkins.io/security/advisory/2022-07-27/#SECURITY-2413)


daniel-beck:
@dillense Thanks for letting us know. We'll confirm the fix and then update the warning.

daniel-beck:
Followed up with request for information in https://github.com/jenkinsci/clif-performance-testing-plugin/pull/13#issuecomment-1213189560.

Kevin-CB:
Hi @dillense,

I'm still able to reproduce the vulnerability, from what I've seen, it's due to [Zip.java#L165](https://github.com/jenkinsci/clif-performance-testing-plugin/blob/7214dfdb32ecd7c23ffb675b4d1e9d419f4fea66/src/main/java/org/ow2/clif/jenkins/jobs/Zip.java#L165).

I'll let you correct it and ping me when it's fixed 🙂

dillense:
Hi @Kevin-CB ,
Strange... I tested and when I import the sample malicious zip, the resulting list is empty, and no file is created (actually I get the Jenkins message "Oops, A problem occurred while processing the request", which is not nice, but safe). Besides, one specific test has been added to the source code.
What is the plug-in version? It should be 70.v7214dfdb_32ec.

dillense:
That said, it is true that there are some "bad" logs in Jenkins. So, it seems the fix is partial (wrt the preview verb).
I need to have a look again.

dillense:
New fix is available in latest release 71.v0741865e206f

Kevin-CB (approved):
I reviewed your fix, it's looks good.

And I see nothing wrong with your pattern!

dillense:
Great! Thanks @Kevin-CB!

dillense:
Hello @daniel-beck . Any update on this? May I give a help?

resources/warnings.json
```diff
@@ -13398,8 +13398,8 @@
     "url": "https://www.jenkins.io/security/advisory/2022-07-27/#SECURITY-2413",
     "versions": [
       {
-        "lastVersion": "64.vc0d66de1dfb_f",
-        "pattern": ".*"
+        "lastVersion": "70.v7214dfdb_32ec",
+        "pattern": "(1[.])|(64[.])|(70[.]).*"
```
daniel-beck, line 13402:
```suggestion
        "pattern": "(1|64|70)(|[.-].+)"
```

daniel-beck:
@dillense Sorry about the delay here.

PR #640, opened 2022-08-15
security fixed in image-tag-parameter 2.0

maduma:
image-tag-paramter 2.0 is using now credential plugin 2.5

### DIFF
--- a/resources/warnings.json
+++ b/resources/warnings.json
@@ -12637,19 +12637,6 @@
     ]
   },
   {
-    "id": "SECURITY-2784-image-tag-parameter",
-    "type": "plugin",
-    "name": "image-tag-parameter",
-    "message": "Stored XSS vulnerability",
-    "url": "https://www.jenkins.io/security/advisory/2022-06-22/#SECURITY-2784",
-    "versions": [
-      {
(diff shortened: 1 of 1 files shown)


daniel-beck:
@maduma Thanks for letting us know. Could you please point me to the specific security fix? Because this issue was never about credentials plugin, I'm confused why you mention that.

Also, please do not delete this entry, instead update the regular expression to `(1)(|[.-].+)` so it only matches vulnerable releases.

maduma:
Hi Daniel,
  
The credential plugin was also affected by the issue. I did not read carrefully the issue and thought updating the credential the dependecy would solve it ... my bad.
  
Do i need to manuany escape all input or de new dependcy to jenkins 2.346.1 will fix the issue ?
  
 
  
Regards
  
-Stephane
  
Daniel Beck wrote:

daniel-beck:
> Do i need to manuany escape all input or de new dependcy to jenkins 2.346.1 will fix the issue ?

Please read https://issues.jenkins.io/browse/SECURITY-2721, which contains a description what you need to change, as well as steps to reproduce the vulnerability (and, once fixed, confirm it's fixed).

PR #644, opened 2022-09-21
Suspend distribution of build-publisher, cons3rt, walti

daniel-beck:
### Build Publisher

The plugin is incompatible with JEP-200, Jenkins 2.266+ (it hardcodes the `j_acegi_security_check` URL), and possibly even always-on CSRF protection (2.222+); so it doesn't really make sense to continue distributing it.

### CONS3RT

The maintainer asked that we suspend distribution with the following explanation:

> The tool it is associated with has moved to a different authentication path than the plugin currently supports, making the plugin no longer valid for use against the tool. As such are no longer looking to support, distribute, or update the plugin.

### Walti

The domain walti.io is for sale, the service seems to no longer exist.


MarkEWaite (approved):

MarkEWaite:
Thanks for doing the cleanup.  Is it OK if I merge this pull request or is it better for you if I leave the merge of pull requests in this repository completely to you?

lsborsne-te:
ok to test

PR #647, opened 2022-09-29
Disambiguate ignore file format

daniel-beck:
Currently it's not possible to easily determine the plugin ID and version being suspended since `-` is both the separator between those values and occurs commonly in both plugin ID and version numbers.

`@` is present in neither (or at least extremely rarely), so makes a better separator.

I tried `:` first, but learned that's an alternative key-value separator in properties files in addition to `=`. While this file could do with some cleanup and adopt a file format similar to the format of `--allowed-artifacts-file` (keys are artifact IDs, values are space separated versions or `*` for all), that would make suspending multiple individual releases of the same plugin for different reasons more cumbersome, so I decided against that for now.

`#` is a comment delimiter, so that's also not great (thanks @zbynek!).


zbynek:
Since `#` is also comment delimiter, maybe there is some even less ambiguous choice. Would `@` work? It's the version delimiter for npm and Rust's Cargo.

daniel-beck:
@zbynek This is great, thanks for the suggestion.

PR #649, opened 2022-10-20
Query repositories in alphabetical order

zbynek:
Fixes https://github.com/jenkins-infra/helpdesk/issues/3190 (when running locally).

Though it is a workaround for a GitHub issue, it may still be worth merging since it makes the code more deterministic.


zbynek:
@daniel-beck do you think this is worth merging? The original issue cannot be reproduced, but deterministic sort may have some upsides.

daniel-beck:
@zbynek It is! Given how relatively infrequent releases of this component are I like to only merge shortly before I go through the test & release cycle.

daniel-beck (approved):

PR #659, opened 2022-12-04
remediate SECURITY-1350 and SECURITY-1351

chonton:
https://github.com/jenkinsci/extended-choice-parameter-plugin/pull/61 
https://github.com/jenkinsci/extended-choice-parameter-plugin/pull/39

### DIFF
--- a/resources/warnings.json
+++ b/resources/warnings.json
@@ -11250,8 +11250,8 @@
     "url": "https://www.jenkins.io/security/advisory/2022-03-15/#SECURITY-1350",
     "versions": [
       {
-        "lastVersion": "359.v35dcfdd0c20d",
-        "pattern": ".*"
+        "lastVersion": " 356.va_90a_94ca_62ec",
+        "pattern": "(0|34[236]|356)(|[.-].+)"
       }
     ]
   },
--- a/resources/warnings.json
+++ b/resources/warnings.json
@@ -11250,7 +11250,7 @@
     "url": "https://www.jenkins.io/security/advisory/2022-03-15/#SECURITY-1350",
     "versions": [
       {
-        "lastVersion": "346.vd87693c5a_86c",
+        "lastVersion": "359.v35dcfdd0c20d",
         "pattern": ".*"
       }
     ]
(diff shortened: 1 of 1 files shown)


resources/warnings.json
```diff
@@ -11263,7 +11263,7 @@
     "url": "https://www.jenkins.io/security/advisory/2022-03-15/#SECURITY-1351",
     "versions": [
       {
-        "lastVersion": "346.vd87693c5a_86c",
+        "lastVersion": "359.v35dcfdd0c20d",
         "pattern": ".*"
```
daniel-beck, line 11267:
```suggestion
        "lastVersion": " 356.va_90a_94ca_62ec",
        "pattern": "(0|34[236]|356)(|[.-].+)"
```

As above.

daniel-beck (dismissed):

resources/warnings.json
chonton: Where is this process documented?  I'd like to prevent myself from making errors in the future.

daniel-beck: Documentation is rather minimal. It hasn't come up a lot, since maintainers can just inform us about fixes, rather than having to submit PRs.

There's Javadoc at https://github.com/jenkinsci/jenkins/blob/8eae71e5464f584533b7482a309ee903e65577f8/core/src/main/java/hudson/model/UpdateSite.java#L833-L843 in core. https://javadoc.jenkins.io/hudson/model/UpdateSite.WarningVersionRange.html is a bit awkward to read because `pattern` is private, so doesn't show, I should probably change that.

Besides that, there are all of the examples in other warnings in this file which give you an idea about what the expected contents are.

chonton:
@daniel-beck I accepted changes

daniel-beck (changes requested):
Per https://github.com/jenkinsci/extended-choice-parameter-plugin/pull/39#issuecomment-1261175575 the warning for SECURITY-1350 remains while the missing permission check isn't added. Manually tested 359.v35dcfdd0c20d, permission checks are still missing.

PR #662, opened 2022-12-07
SECURITY-2617-extended-choice-parameter & SECURITY-2232

chonton:
resolved by https://github.com/jenkinsci/extended-choice-parameter-plugin/commit/5f703277898c801642a1755a6371a1c234341eaf

### DIFF
--- a/resources/warnings.json
+++ b/resources/warnings.json
@@ -11367,8 +11367,8 @@
     "url": "https://www.jenkins.io/security/advisory/2022-03-15/#SECURITY-2232",
     "versions": [
       {
-        "lastVersion": "346.vd87693c5a_86c",
-        "pattern": ".*"
+        "lastVersion": "0.56",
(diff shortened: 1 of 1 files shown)


daniel-beck:
@chonton Could you provide some context for this? The linked commit pre-dates the advisories by six years, and we confirmed exploitability in the "last version".

resources/warnings.json
```diff
@@ -11367,8 +11367,8 @@
     "url": "https://www.jenkins.io/security/advisory/2022-03-15/#SECURITY-2232",
     "versions": [
       {
-        "lastVersion": "346.vd87693c5a_86c",
-        "pattern": ".*"
+        "lastVersion": "0.56",
+        "pattern": "0\\.\\d\\d?"
```
daniel-beck, line 11371: This makes no sense, because `0.56` is not the last release with major version `0`, so the pattern matches later versions.

PR #668, opened 2022-12-21
Deprecate unleash plugin

MarkEWaite:
## Deprecate unleash plugin

https://github.com/shillner/unleash-maven-plugin/ has been archived.  That is the unleash plugin for Apache Maven.  The Jenkins unleash maven plugin depends on the unleash plugin for Apache Maven.

https://github.com/jenkinsci/unleash-plugin has been archived.  That is the Jenkins unleash Maven plugin. The last release of the Jenkins unleash maven plugin was 3 years ago. Since the Jenkins plugin repository has been archived, we can expect no new releases of the plugin.

[JENKINS-70326](https://issues.jenkins.io/browse/JENKINS-70326) notes that the Jenkins unleash maven plugin does not work with Java 17.

Let's deprecate the Jenkins plugin so that it is clear to users that the Jenkins plugin is not being maintained and that the Apache Maven plugin is not being maintained.

The [deprecation document](https://www.jenkins.io/doc/developer/plugin-governance/deprecating-or-removing-plugin/) says that the preferred way to deprecate a plugin is to add the "deprecated" topic to the repository and to explain in the plugin documentation why it is being deprecated.  That would mean allowing writes to the repository again, just long enough to set the topic.

I assume that is more complicated than the alternate method, a pull request to the update center repository to mark the repository as deprecated.  I've taken the alternate approach with this pull request.

https://github.com/jenkins-infra/plugins-wiki-docs/pull/6 is the documetnation pull request to mark the plugin as deprecated.


wborn (approved):
Thanks for making all the PRs!

PR #670, opened 2023-01-05
fix: Support web components for nav and footer on jenkins.io

halkeye:
https://github.com/jenkins-infra/jenkins.io/pull/5668#issuecomment-1371927933

makes sure property is set right for components
also don't include report a problem / improve this page (but easy to add if desired)

### DIFF
--- a/site/generate-jenkins-io-template.sh
+++ b/site/generate-jenkins-io-template.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+WWW_ROOT_DIR="${WWW_ROOT_DIR:-${1:-./www2}}"
+SELF_URL="${SELF_URL:-${2:-https://updates.jenkins.io}}"
+
+wget -q --convert-links -O "$WWW_ROOT_DIR/index.html" --convert-links https://www.jenkins.io/templates/updates/index.html
+sed -i -e "s|<jio-navbar.*/jio-navbar>|<jio-navbar class='fixed-top' id='ji-toolbar' property='${SELF_URL}'></jio-navbar>|g" "$WWW_ROOT_DIR/index.html"
+sed -i -e "s|<jio-footer.*/jio-footer>|<jio-footer property='${SELF_URL}'></jio-footer>|g" "$WWW_ROOT_DIR/index.html"
--- a/site/generate.sh
+++ b/site/generate.sh
@@ -165,5 +165,5 @@ popd
 
 # copy other static resource files
 echo '{}' > "$WWW_ROOT_DIR/uctest.json"
-wget -q --convert-links -O "$WWW_ROOT_DIR/index.html" --convert-links https://www.jenkins.io/templates/updates/index.html
+"${MAIN_DIR}/site/generate-jenkins-io-template.sh" "$WWW_ROOT_DIR" "https://updates.jenkins.io"
 cp -av "tmp/tiers.json" "$WWW_ROOT_DIR/tiers.json"


site/generate-jenkins-io-template.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+WWW_ROOT_DIR="${WWW_ROOT_DIR:-${1:-./www2}}"
+SELF_URL="${SELF_URL:-${2:-https://updates.jenkins.io}}"
+
+wget -q --convert-links -O "$WWW_ROOT_DIR/index.html" --convert-links https://www.jenkins.io/templates/updates/index.html
```
daniel-beck, line 6: Please use long options in scripts where possible (e.g. `sed` isn't for compatibility) to make them easier to understand.

```suggestion
wget --quiet --convert-links --output-document "$WWW_ROOT_DIR/index.html" --convert-links https://www.jenkins.io/templates/updates/index.html
```

site/generate-jenkins-io-template.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+WWW_ROOT_DIR="${WWW_ROOT_DIR:-${1:-./www2}}"
+SELF_URL="${SELF_URL:-${2:-https://updates.jenkins.io}}"
```
daniel-beck, line 4: Please add a comment to the beginning of the script that explains usage.

site/generate-jenkins-io-template.sh
```diff
@@ -0,0 +1,8 @@
...
+
+wget -q --convert-links -O "$WWW_ROOT_DIR/index.html" --convert-links https://www.jenkins.io/templates/updates/index.html
+sed -i -e "s|<jio-navbar.*/jio-navbar>|<jio-navbar class='fixed-top' id='ji-toolbar' property='${SELF_URL}'></jio-navbar>|g" "$WWW_ROOT_DIR/index.html"
+sed -i -e "s|<jio-footer.*/jio-footer>|<jio-footer property='${SELF_URL}'></jio-footer>|g" "$WWW_ROOT_DIR/index.html"
```
daniel-beck, line 8: This is incompatible with Mac OS `sed` which does not allow for a parameter-less `-i`.
```suggestion
sed -i.bak -e "s|<jio-navbar.*/jio-navbar>|<jio-navbar class='fixed-top' id='ji-toolbar' property='${SELF_URL}'></jio-navbar>|g" "$WWW_ROOT_DIR/index.html"
sed -i.bak -e "s|<jio-footer.*/jio-footer>|<jio-footer property='${SELF_URL}'></jio-footer>|g" "$WWW_ROOT_DIR/index.html"
```

(Plus I recommend to `rm` the backup file or adding to `.gitignore`)

daniel-beck, line 8: Why is this a separate `sed` call? Can't this be an additional expression in the first `sed` call? Readability perhaps?

site/generate-jenkins-io-template.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+WWW_ROOT_DIR="${WWW_ROOT_DIR:-${1:-./www2}}"
+SELF_URL="${SELF_URL:-${2:-https://updates.jenkins.io}}"
+
```
daniel-beck, line 5:
```suggestion

set -o errexit
set -o nounset
set -o pipefail

```

daniel-beck (commented):
Thanks for the adaptation.

site/generate-jenkins-io-template.sh
halkeye: yea i guess, it never really occured to me to join them, i found it easier to read i guess?

site/generate-jenkins-io-template.sh
```diff
@@ -0,0 +1,12 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+wget --quiet --convert-links --output-document "$WWW_ROOT_DIR/index.html" --convert-links https://www.jenkins.io/templates/updates/index.html
+sed -i.bak -e "s|<jio-navbar.*/jio-navbar>|<jio-navbar class='fixed-top' id='ji-toolbar' property='${SELF_URL}'></jio-navbar>|g" "$WWW_ROOT_DIR/index.html"
```
halkeye, line 13: does mac allow `-i''`? There's no need for a backup since the file is pulled down seconds earlier
would it be better to switch to `perl -pi -e`, i know that works almost everwhere.

Honestly it probably could just be pipes, output to stdout, then pipe to sed (optionally two seds) then write to file

daniel-beck: `-i ''` (note the space, `-i''` is just the same as `-i`) would work on Mac OS, but not Linux.

site/generate-jenkins-io-template.sh
daniel-beck: Thanks for adding the doc, but by usage I meant something like https://github.com/jenkins-infra/update-center2/blob/c5841e67ed066988bf4c3ce671fc2c1c95b3382a/site/generate.sh#L3

(Since the other script cannot be used without args it's part of the default invocation, that cannot be done here of source. Check for `$# -gt 2` might still be a useful error condition.

halkeye:
@daniel-beck it'll be at least a day before i can poke at this again (work then personal meeting), do you want to make the changes you requested yourself?

otherwise i'll try to fix things up after everything tonight, but likely not till tomorrow afternoon for me

daniel-beck:
> it'll be at least a day before i can poke at this again (work then personal meeting), do you want to make the changes you requested yourself?
> 
> otherwise i'll try to fix things up after everything tonight, but likely not till tomorrow afternoon for me

I don't think I'll have time before then. Thanks for checking!

halkeye:
Cleanup generate jenkins io template                                                                                                                          
  
  * Switch to cat | sed to make it more osx portable
  * Document with usage when bad parameters

I opted out of having Improve This Page and Report a Problem links on the footer but they are simple html attributes if we want them.

daniel-beck (changes requested):
https://github.com/jenkins-infra/update-center2/blob/master/src/main/java/io/jenkins/update_center/JenkinsIndexTemplateProvider.java needs to be changed as well for pages like https://updates.jenkins.io/download/war/ or https://updates.jenkins.io/download/plugins/matrix-auth/

PR #677, opened 2023-01-24
Deprecate handlebars plugin

daniel-beck:
Handlebars 3 is [abandoned](https://github.com/handlebars-lang/handlebars.js/tree/3.x), [has unresolved vulnerabilities](https://github.com/handlebars-lang/handlebars.js/issues/1736#issuecomment-836430832), and [no other plugin depends on this one anymore](https://plugins.jenkins.io/handlebars/#dependencies).

So far only marking it as deprecated instead of suspending distribution, but could go either way. CasC would probably be impacted by the latter?

Maintainers @batmat @rsandell to approve this PR.


rsandell (approved):

MarkEWaite (approved):

batmat (approved):

lambertJ9:
what is the solution

PR #686, opened 2023-03-01
Further limit allowed characters in file path

daniel-beck:
Amends #364 and https://github.com/jenkins-infra/update-center2/pull/673.

This filters out paths that include a bunch of characters that indicate something has gone wrong during the Maven release process. While there are already several characters (notably unprintable ASCII, originally motivating this filter) being filtered, this adds more, based on what is actually used: Keep basic ASCII letters, digits, `_+.-`, as well as `/` since we're looking at paths at this point.

Compared to the current filter, this doesn't remove anything that is being distributed at the moment.

Also fix the message, it hasn't been correct in a long time, most of the stuff filtered is due to `%` from percent-encoded characters.

### DIFF
--- a/src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
+++ b/src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
@@ -79,7 +79,12 @@ public class ArtifactoryRepositoryImpl extends BaseMavenRepository {
     }
 
     private static boolean containsIllegalChars(String test) {
-        return !test.chars().allMatch(c -> c >= 0x2B && c < 0x7B);
+        return !test.chars().allMatch(c ->
+                c >= 0x30 && c <= 0x39 // allow digits 0-9
+                || c >= 0x41 && c <= 0x5A // allows letter A-Z
+                || c >= 0x61 && c <= 0x7A // allows letter a-z
+                || c == 0x2B || c >= 0x2D && c <= 0x2F || c == 0x5F // allows: +-./_
+        );
     }
 
     private static ArtifactCoordinates toGav(JsonFile f) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
```diff
@@ -79,15 +79,20 @@ protected Set<ArtifactCoordinates> listAllJenkinsWars(String groupId) throws IOE
     }
 
     private static boolean containsIllegalChars(String test) {
-        return !test.chars().allMatch(c -> c >= 0x2B && c < 0x7B);
+        return !test.chars().allMatch(c ->
+                c >= 0x30 && c <= 0x39 // allow digits 0-9
+                || c >= 0x41 && c <= 0x5A // allows letter A-Z
+                || c >= 0x61 && c <= 0x7A // allows letter a-z
+                || c == 0x2B || c >= 0x2D && c <= 0x2F || c == 0x5F // allows: +-./_
```
zbynek, line 86: Looks good, but why do you use hex codes?
```suggestion
        return !test.chars().allMatch(c ->
                c >= '0' && c <= '9'
                || c >= 'A' && c <= 'Z'
                || c >= 'a' && c <= 'z'
                || c == '+' || c == '-' || c <= '.' || c == '/' || c == '_'
```

zbynek (approved):

src/main/java/io/jenkins/update_center/ArtifactoryRepositoryImpl.java
daniel-beck: This went through a few iterations, and started by having range boundaries that made no sense as characters (e.g. `+` through `9` or `_` through `z`), and I didn't adapt the representation when I went with more easily consumable ranges.

NotMyFault (approved):

PR #691, opened 2023-03-17
Ignore h2-api@10.v8dd81336a_ede

timja:
See https://github.com/jenkinsci/h2-api-plugin/pull/9

Is it possible to fix this? Given it's been pushed to the maven repository?
i.e. renovate / dependabot might just keep offering the new version?

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -833,3 +833,6 @@ ascii-magician
 
 # https://github.com/jenkinsci/gitlab-plugin/issues/1419
 jersey2-api@2.39-1
+
+# https://github.com/jenkinsci/h2-api-plugin/pull/9
+h2-api@10.v8dd81336a_ede


daniel-beck:
Anyone who's already installed this version would never receive another update.

The way forward it to go with a top-level version greater than 10.

For now I'm merging this in case someone has a fantastic idea, but I think the nice versions for this plugin are now burned. I recommend being more careful in the future, e.g. the command lines from https://www.jenkins.io/doc/developer/publishing/releasing-cd/#update-maven-pom-and-config can be used to ensure the version looks as expected.

timja:
> @timja Any news here? Or are we waiting for monthly usage stats?

I'm thinking of just hardcoding a prefix of 11 unless any better ideas

daniel-beck:
https://stats.jenkins.io/pluginversions/h2-api.html has 2.5% of all installs.

timja:
Interesting, thanks should be fixed by https://github.com/jenkinsci/h2-api-plugin/pull/9

PR #695, opened 2023-03-30
Discontinuing the Delphix plugin

nick-mathison:
The Delphix plugin is out of date and requires significant attention. Our goal is to temporarily discontinue its distribution until it's rewritten to our higher standards. 

It's my understanding that this page will become unavailable on merge: https://plugins.jenkins.io/delphix/

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -73,6 +73,7 @@ cppunit = https://wiki.jenkins-ci.org/display/JENKINS/CppUnit+Plugin
 create-fingerprint-plugin        # renamed to create-fingerprint
 datadog-build-reporter           # renamed to datadog
 delphix-plugin                   # renamed to delphix
+delphix                          # discontinued
 description-column               # renamed to description-column-plugin
 discobit-autoconfig = https://github.com/jenkins-infra/update-center2/pull/587
 # removal requested by teilo, superceded by dockerhub-notification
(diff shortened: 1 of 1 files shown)


resources/artifact-ignores.properties
```diff
@@ -73,6 +73,7 @@ cppunit = https://wiki.jenkins-ci.org/display/JENKINS/CppUnit+Plugin
 create-fingerprint-plugin        # renamed to create-fingerprint
 datadog-build-reporter           # renamed to datadog
 delphix-plugin                   # renamed to delphix
+delphix                          # discontinued
```
daniel-beck, line 76:
```suggestion
delphix = https://github.com/jenkins-infra/update-center2/pull/695
```

daniel-beck (commented):
> It's my understanding that this page will become unavailable on merge: https://plugins.jenkins.io/delphix/

That's correct, however we prefer to have a reference URL documenting this state. Could be as simple as the URL to this pull request. That marks the plugin additionally as deprecated and informs current users. See suggested change.

Since you're not a committer of this plugin, please have @mcred approve this request here, or `mcred` or `arunskurian` via Jira.

mcred:
I just chatted with @arunskurian and this is approved.

daniel-beck:
@cpwr-devops Nobody mentioned you so my best guess is you've subscribed to notifications from this repo. You can unsubscribe yourself.

PR #699, opened 2023-05-02
Remove security warning (SECURITY-835) from plugin 

nikhil-dabhade:
The issue SECURITY-835 was resolved in the plugin repo on Aug 6th, 2019 with this code commit:
2d8cfe3cdd299c152ca5d13aada51cb69d0fec88

The plugin has been released multiple times since then.

Link to plugin repo and the commit that resolved the issue:
https://github.com/jenkinsci/aws-device-farm-plugin/commit/2d8cfe3cdd299c152ca5d13aada51cb69d0fec88

### DIFF
--- a/resources/warnings.json
+++ b/resources/warnings.json
@@ -15634,5 +15634,213 @@
         "pattern": "([1-3]|4[.][0-4]|4[.]5[.][0-2])(|[.-].+)"
       }
     ]
+  },
+  {
+    "id": "SECURITY-2837",
+    "type": "plugin",
+    "name": "thycotic-secret-server",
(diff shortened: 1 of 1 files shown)


nikhil-dabhade:
Hi, Please let me know if there is anything else needed on this pull request.

zbynek:
This would remove the warning entirely, even for users who still have an older version.
Instead of removing,  `pattern` should be changed to match all the affected versions and `lastVersion` should be specified (the last affected version, not the first fixed one).

nikhil-dabhade:
Makes sense. Thank you for the feedback. I will get that in today.

PR #700, opened 2023-05-03
HttpsHelper_optimised

rishav887:
The changes I made include:

1) Adding a check for null response body in the body() method and throwing an IllegalStateException if the body is null.

2) Adding a try-catch block in the getResponseBody() method to handle exceptions more robustly. The catch block first re-throws any IOException or IllegalStateException that may have occurred, and then catches all other exceptions and wraps them in an IOException with a more descriptive error message.

### DIFF
--- a/src/main/java/io/jenkins/update_center/util/Environment.java
+++ b/src/main/java/io/jenkins/update_center/util/Environment.java
@@ -8,39 +8,41 @@ import java.util.logging.Logger;
 public final class Environment {
     private Environment() {}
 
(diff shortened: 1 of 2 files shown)


zbynek:
Does this solve any actual problem? 

If you're just looking for a way to contribute to Jenkins, you can check https://www.jenkins.io/participate/code/ or the list of good first issues https://github.com/search?q=org%3Ajenkinsci+org%3Ajenkins-infra+is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22&type=issues

rishav887:
sure, thanks

PR #703, opened 2023-05-14
Suspend distribution of `whitesource` plugin

ChenLuigi:


daniel-beck (changes requested):

daniel-beck:
@ChenLuigi Thanks for letting us know. I've suggested a fix and improvement so current plugin users can be more clearly informed about this.

FTR @ChenLuigi is a committer to the plugin, so this request to suspend looks fine.

daniel-beck (approved):

PR #704, opened 2023-05-18
add credentials to the allowed topics

jtnord:
many plugins provide credential types or are credential stores.   if they are labeled as such we should have an easy way to discover them.

as per https://www.jenkins.io/doc/developer/publishing/documentation/#labeling-plugins submitting a PR to add `credentials` to the allow list of labels.

Not sure where the plugin page gets the text labels to search specific labels from in order to change that too :-/

### DIFF
--- a/resources/allowed-github-topics.properties
+++ b/resources/allowed-github-topics.properties
@@ -24,6 +24,7 @@ cmp
 codebuild
 confluence
 configuration-as-code
+credentials
 database
 deployment
 devops


daniel-beck:
> Not sure where the plugin page gets the text labels to search specific labels from in order to change that too :-/

https://github.com/jenkins-infra/plugin-site-api/blob/416279518ff3444904c28b1ef3aa56ca3ff7d38b/src/main/resources/categories.json / https://github.com/jenkins-infra/plugin-site-api/blob/416279518ff3444904c28b1ef3aa56ca3ff7d38b/src/main/resources/labels.json ?

zbynek:
Jenkins core properties files are used via https://github.com/jenkins-infra/plugin-site/blob/a06c912a5371e87254d6b88eeb19a974bd947292/plugins/gatsby-source-jenkinsplugins/utils.js#L380 for translations and https://github.com/jenkins-infra/plugin-site/blob/a06c912a5371e87254d6b88eeb19a974bd947292/plugins/gatsby-source-jenkinsplugins/categories.json for the category tree. The `plugin-site-api` repo is mostly dead code now.

PR #678, opened 2023-01-24
Suspend keycloak plugin

Kevin-CB:
We have been notified by the maintainers that this plugin is no longer maintained and that they wish to depublish it.

PR #706, opened 2023-05-22
Revert #678 to reinstate keycloak plugin.

mrmateo:
I have adopted the plugin so it now as an active maintainer and version 2.3.1 was released that fixes security vulnerabilities SECURITY-2987 and SECURITY-2986 .


daniel-beck:
Thanks for letting us know.

Unlikely when a plugin is suspended _because_ of its security issues, we can merge this now and update the security warnings (including confirming that issues are resolved) in a second step.

daniel-beck (approved):

PR #708, opened 2023-05-30
Deprecate and suspend `thundra-foresight`

suleymanbarman:
thundra-foresight plugin is no longer in service


daniel-beck:
@suleymanbarman Thanks for letting us know!

PR #711, opened 2023-06-12
Suspend distribution of `config-autorefresh-plugin`

TobiX:
The related core feature was removed in 2.223, so this plugin does
nothing.

(Opening a PR here seems to be the easier way then to adopt & then just
deprecate th plugin :P)


NotMyFault:
> The related core feature was removed in 2.223, so this plugin does
> nothing.

Wouldn't it be better to suspend the plugin then? If the feature doesn't exist in modern Jenkins, there's no reason to install the plugin.

daniel-beck:
@TobiX Do you plan to amend this PR? If not, would you object to me repurposing it for (or replacing it with another) suspending the plugin?

TobiX:
@daniel-beck Go ahead, currently quite busy with other stuff 😢

daniel-beck:
@NotMyFault WDYT?

MarkEWaite:
+1 from me to suspend distribution rather than mark it as deprecated

PR #716, opened 2023-06-24
Suspend blackduck-detect plugin 8.0.0 due to multiple serious unresolved issues

MarkEWaite:
## Suspend blackduck-detect plugin 8.0.0

The Synopsys Detect plugin 8.0.0 release that was delivered in Jan 2023 has several issues that are unresolved.  Those issues are severe enough to justify suspension of delivery of that release of the plugin.

* 8.0.0 crashes startup of Jenkins 2.410 and later (June 2023)
  https://issues.jenkins.io/browse/JENKINS-71480

* 8.0.0 includes far too many dependencies (including Guava 23, jackson 2.10, etc.) (Apr 2023)
  https://issues.jenkins.io/browse/JENKINS-70671

* 8.0.0 bundles pull-parser.jar by mistake (Feb 2023)
  https://issues.jenkins.io/browse/JENKINS-71023

The blackduck-detect plugin 7.0.0 does not crash Jenkins 2.410 and seems unlikely to have the other issues either.

CC: @sig-cpotts, @jppetrakis, @dmamidisynopsys, @DanaMaxfield

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -847,3 +847,9 @@ compuware-scm-downloader@2.0.16
 # https://jenkins.io/security/advisory/2023-03-21/
 mashup-portlets-plugin = https://www.jenkins.io/security/plugins/#suspensions
 remote-jobs-view-plugin = https://www.jenkins.io/security/plugins/#suspensions
+
+# Suspend blackduck-detect 8.0.0 plugin due to multiple serious unresolved issues
+# https://issues.jenkins.io/browse/JENKINS-71480 - 8.0.0 crashes startup of Jenkins 2.410 and later (June 2023)
+# https://issues.jenkins.io/browse/JENKINS-70671 - 8.0.0 includes far too many dependencies (including Guava 23, jackson 2.10, etc.) (Apr 2023)
+# https://issues.jenkins.io/browse/JENKINS-71023 - 8.0.0 bundles pull-parser.jar by mistake (Feb 2023)
+blackduck-detect@8.0.0


resources/artifact-ignores.properties
```diff
@@ -847,3 +847,9 @@ compuware-scm-downloader@2.0.16
 # https://jenkins.io/security/advisory/2023-03-21/
 mashup-portlets-plugin = https://www.jenkins.io/security/plugins/#suspensions
 remote-jobs-view-plugin = https://www.jenkins.io/security/plugins/#suspensions
+
+# Suspend blackduck-detect 8.0.0 plugin due to multiple serious unresolved issues
+# https://issues.jenkins.io/browse/JENKINS-71480 - 8.0.0 crashes startup of Jenkins 2.410 and later (June 2023)
+# https://issues.jenkins.io/browse/JENKINS-70671 - 8.0.0 includes far too many dependencies (including Guava 23, jackson 2.10, etc.) (Apr 2023)
+# https://issues.jenkins.io/browse/JENKINS-71023 - 8.0.0 bundles pull-parser.jar by mistake (Feb 2023)
```
NotMyFault, line 854:
```suggestion
# https://issues.jenkins.io/browse/JENKINS-71023 - 8.0.0 bundles pull-parser.jar by mistake (Feb 2023)
# https://github.com/jenkinsci/synopsys-detect-plugin/pull/53 - Bundling the Jenkins core jar and plugins
```

NotMyFault (approved):

Wadeck (approved):
Thanks Mark for the list. It seems reasonable 👍

zbynek:
I wonder if all of the issues are just symptoms of the same problem -- having 2 different versions of both Jenkins Core and Guava on the classpath is bound to backfire. https://github.com/jenkinsci/synopsys-detect-plugin/pull/53

MarkEWaite:
> I wonder if all of the issues are just symptoms of the same problem -- having 2 different versions of both Jenkins Core and Guava on the classpath is bound to backfire. [jenkinsci/synopsys-detect-plugin#53](https://github.com/jenkinsci/synopsys-detect-plugin/pull/53)

I suspect that they are all symptoms of the same problem.  At the root, I think the plugin needs to switch from gradle to maven and that is likely to resolve most of the issues.  However, without the involvement of a maintainer, no change will happen and more and more Jenkins users will be negatively affected by this plugin.

dmamidisynopsys:
@MarkEWaite Hi Mark, thanks for reaching out. We are actively investigating the 3 issues reported. Can you share any logs which points us that crash issue is caused by Detect Jenkins plugin for Jenkins 2.410 and later? We have fixes for two other issues reported that we are testing and validating before we can release. The fixes to issues will be released in 8.0.1 version of Detect plugin. Apologies for the inconvenience. We will provide you an update soon.

MarkEWaite:
> @MarkEWaite Hi Mark, thanks for reaching out. We are actively investigating the 3 issues reported. Can you share any logs which points us that crash issue is caused by Detect Jenkins plugin for Jenkins 2.410 and later?

A series of steps to duplicate the problem are included in [JENKINS-71480](https://issues.jenkins.io/browse/JENKINS-71480).  The issue report includes the stack trace.  I suspect that the outdated version of Guava that is packaged with the 8.0.0 plugin releases is the most likely cause of the Jenkins crash during startup.

dmamidisynopsys:
Hi Mark,

Yes, we are trying to use those steps to reproduce. On the 4th step, when we restart jenkins, the .sh process gets killed for both cases when Detect plugin is installed and not installed. Can you recommend how do we perform the 4th step of restart to be able to reproduce the issue when plugin is installed ? Thanks!


> > @MarkEWaite Hi Mark, thanks for reaching out. We are actively investigating the 3 issues reported. Can you share any logs which points us that crash issue is caused by Detect Jenkins plugin for Jenkins 2.410 and later?
> 
> A series of steps to duplicate the problem are included in [JENKINS-71480](https://issues.jenkins.io/browse/JENKINS-71480). The issue report includes the stack trace. I suspect that the outdated version of Guava that is packaged with the 8.0.0 plugin releases is the most likely cause of the Jenkins crash during startup.

dmamidisynopsys:
@MarkEWaite Another question - When you say you are suspecting outdating version of Guava might be the problem. Guava transitive dependency is in fact coming from org.jenkins-ci.main:jenkins-core:2.377, not by direct reference by the plugin. Are you suggesting we exclude that one ?

MarkEWaite:
> @MarkEWaite Another question - When you say you are suspecting outdating version of Guava might be the problem. Guava transitive dependency is in fact coming from org.jenkins-ci.main:jenkins-core:2.377, not by direct reference by the plugin. Are you suggesting we exclude that one ?

Don't worry about my guess.  I'm probably wrong.  I frequently make poor guesses that are later proven wrong.

If the contents of the new hpi file have been reduced to stop including Jenkins core and stapler and winp and xstream and spring core and Guava and groovy and 10+ Jenkins plugins, and if the new hpi file works as expected in Jenkins 2.411, then my poor guess should be ignored.

The 8.0.0 hpi file includes over 140 jar files.

dmamidisynopsys:
We were already testing the theory of having two different versions of jenkins-core, guava and other few dependencies causing the issue. Removing them seems to have fixed the issue. We will need to validate and run all our regressions tests before we can release 8.0.1. Don't think your guess is wrong :) 

We are making similar changes as https://github.com/jenkinsci/synopsys-detect-plugin/pull/53 


> > @MarkEWaite Another question - When you say you are suspecting outdating version of Guava might be the problem. Guava transitive dependency is in fact coming from org.jenkins-ci.main:jenkins-core:2.377, not by direct reference by the plugin. Are you suggesting we exclude that one ?
> 
> Don't worry about my guess. I'm probably wrong. I frequently make poor guesses that are later proven wrong.
> 
> If the contents of the new hpi file have been reduced to stop including Jenkins core and stapler and winp and xstream and spring core and Guava and groovy and 10+ Jenkins plugins, and if the new hpi file works as expected in Jenkins 2.411, then my poor guess should be ignored.
> 
> The 8.0.0 hpi file includes over 140 jar files, including all those files that I listed earlier. That's too much in the hpi file.

MarkEWaite:
> Hi Mark,
> 
> Yes, we are trying to use those steps to reproduce. On the 4th step, when we restart jenkins, the process gets killed with this exception "./run-jenkins.sh: line 26: 83445 Killed: 9 JENKINS_HOME=. java -jar ../$JENKINS_WAR" for both cases when Detect plugin is installed and not installed. Can you recommend how do we perform the 4th step of restart to be able to reproduce the issue when plugin is installed ? Thanks!

Be sure that each run of the `run-jenkins.sh` script starts from a directory that contains only `run-jenkins.sh` and `plugins.txt`.  If the `plugins.txt` file includes the line `blackduck-detect:8.0.0` then Jenkins will fail to start.  If the `plugins.txt` file does not include the line `blackducj-detect:8.0.0`, then Jenkins will start successfully.

daniel-beck:
> "./run-jenkins.sh: line 26: 83445 Killed: 9 JENKINS_HOME=. java -jar ../$JENKINS_WAR" for both cases when Detect plugin is installed and not installed.

Might be a known issue with M1 Macs failing to restart; should be started again manually.

daniel-beck:
Given the maintainer response, is this PR obsolete?

MarkEWaite:
> Given the maintainer response, is this PR obsolete?

Yes.  Thanks for everyone's attention.  I think that the `blackduck-detect` plugin 8.0.0 is toxic to Jenkins users because it crashes newer Jenkins versions on startup.  However, so long as a repaired version is released promptly, I don't believe there is enough benefit to justify suspending the distribution of the 8.0.0 release.

dmamidisynopsys:
Thanks @MarkEWaite. Our ETA of 8.0.1 version release is end of today. We will make a community announcement and send out heads up emails, release notes to our customers/users.

MarkEWaite:
I see that 8.0.1 has been released to the Jenkins artifact repository and to the [Jenkins update center](https://updates.jenkins.io/latest/blackduck-detect.hpi) and that it is visible on [at least two of the mirrors](https://get.jenkins.io/plugins/blackduck-detect/8.0.1/blackduck-detect.hpi?mirrorlist).  I've downloaded from a mirror and confirmed that the file matches with the file available from the artifact repository.

I am surprised that the 8.0.1 release includes `WEB-INF/lib/guava-23.3-jre.jar`, but hope that you've already confirmed that the inclusion of that outdated copy of guava does not harm the plugin and does not harm Jenkins.

PR #717, opened 2023-06-25
Update instructions of artifact-ignores.properties

damianszczepanik:
Use proper separator in readme section


daniel-beck:
Thanks!

PR #718, opened 2023-06-25
Remove invalid sidebar-link releases

damianszczepanik:
Artifacts were released only for testing fix https://github.com/jenkinsci/sidebar-link-plugin/pull/60 and should not be listed in https://plugins.jenkins.io/sidebar-link/releases/ as new valid release is published already.

New artifact is not listed as the most recent version so that's why it is not available at update center.


damianszczepanik:
@daniel-beck ready for review

daniel-beck:
Should be gone from https://updates.jenkins.io/download/plugins/sidebar-link/ in a few minutes. plugins.jenkins.io probably takes longer.

PR #722, opened 2023-06-30
Remove warning for SECURITY-3013

tdraebing:
The security issue was resolved in release 398.v3dfa_cb_223984 of the lucene-search plugin.


daniel-beck:
Thanks, I've merged #721 instead.

PR #725, opened 2023-07-06
Re-releasing the Delphix Plugin

nick-mathison:
We plan to push/release an updated version of our Delphix plugin v3.0.0 tomorrow. 
This will replace the existing (but currently hidden) Delphix Plugin. 

This PR is meant to revert the change I made back in March. https://github.com/jenkins-infra/update-center2/pull/695

Ideally, we merge this PR around the same time [the PR here is merged. ](https://github.com/jenkinsci/delphix-plugin)

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -861,17 +861,17 @@ synopsys-polaris@1.3.3
 
 # Removing old Delphix Plugin releases
 # https://github.com/jenkins-infra/update-center2/pull/725
-delphix-plugin@1.0.0
-delphix-plugin@1.0.1
-delphix-plugin@1.0.2
-delphix-plugin@1.0.3
-delphix-plugin@1.0.4
-delphix-plugin@1.0.5
-delphix-plugin@1.0.6
-delphix-plugin@1.0.7
-delphix-plugin@1.0.8
-delphix-plugin@2.0.0
-delphix-plugin@2.0.1
-delphix-plugin@2.0.2
-delphix-plugin@2.0.3
-delphix-plugin@2.0.4
+delphix@1.0.0
+delphix@1.0.1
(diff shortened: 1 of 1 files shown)


daniel-beck:
Please note this will make earlier versions of the plugin also available for download, in particular they'll be served to anyone running Jenkins older than 2.361.4 (the minimum for the updated plugin). If the intention is to not distribute them again, those old versions need to be listed individually.

Also, what's the experience upgrading from an earlier release like? The description of PR 20 reads like it's basically a different plugin with the same ID, if so, likely resulting in upgrades losing all of the data? We generally discourage that, as it goes against [our values](https://www.jenkins.io/project/governance/#compatibility-matters). The PR even removes the prior `compatibleSinceVersion` declaration entirely instead of changing it to `3.0`, so users are not even warned before updating.

nick-mathison:
Separately, we are having trouble merging our Jenkins branch which I suspect is a result of not having permissions. 
https://github.com/jenkins-infra/repository-permissions-updater/issues/3426
Your confirmation would be helpful. Thanks!

daniel-beck:
> Please let us know how you'd like us to move forward. We believe this is the final item that is blocking our release.

Well, and the lack of a release, no? https://repo.jenkins-ci.org/releases/org/jenkins-ci/plugins/delphix/ has nothing. You uploaded a snapshot, but that's it.

Note that this does not block releases, it just prevents them from being available on update sites for download by users.

ankit-patil-hubs:
@daniel-beck As suggested by you we have released a new version 3.0.1 , same is available on [https://repo.jenkins-ci.org/releases/org/jenkins-ci/plugins/delphix/](https://repo.jenkins-ci.org/releases/org/jenkins-ci/plugins/delphix/) for unblocking the PR merge.

PR #728, opened 2023-07-18
Remove SECURITY-1450 for v3+ releases - Updated regex pattern

nick-mathison:
Updating Regex Pattern to remove the SECURITY-1450 vulnerability. 
https://www.jenkins.io/security/advisory/2019-10-16/#SECURITY-1450

This issue has been replaced in the new Delphix v3.0.1 release. 

We now leverage Secret text to authenticate via API Keys.

The new regex should target all pre-v3 releases.

### DIFF
--- a/resources/warnings.json
+++ b/resources/warnings.json
@@ -6137,7 +6137,7 @@
     "versions": [
       {
         "lastVersion": "2.0.4",
-        "pattern": "(^[012].*)"
+        "pattern": "([12])(|[.-].*)"
       }
     ]
   },
--- a/resources/warnings.json
+++ b/resources/warnings.json
@@ -6137,7 +6137,7 @@
     "versions": [
       {
         "lastVersion": "2.0.4",
-        "pattern": ".*"
+        "pattern": "(^[012].*)"
       }
     ]
   },


arunskurian (approved):

daniel-beck:
Thanks for letting us know. We'll confirm this and then merge.

resources/warnings.json
```diff
@@ -6137,7 +6137,7 @@
     "versions": [
       {
         "lastVersion": "2.0.4",
-        "pattern": ".*"
+        "pattern": "(^[012].*)"
```
daniel-beck, line 6140:
```suggestion
        "pattern": "([12])(|[.-].*)"
```

Otherwise will match version `10`. Other changes for consistency.

nick-mathison: Hi Daniel, 

Thanks for the updates. My RegEx capabilities were never great, so I appreciate you checking the string. 

Can you confirm this will work from your end? I'm quick RegEx101 test is unable to identify the entire string. Thanks!

daniel-beck: > I'm quick RegEx101 test is unable to identify the entire string

The "match" view on that site seems to have `#find` semantics rather than `#matches`. The "unit test" view seems more suitable for this use case.

I confirm the format of regexes we've been using since about 2016 in this file works 😉

nick-mathison:
Hi @daniel-beck, please merge this PR if there aren't any other issues. Thanks!

Kevin-CB:
I can confirm that the vulnerability reported in SECURITY-1450 is no longer an issue, as the feature was removed in v3.0.1.

PR #730, opened 2023-07-31
Remove SECURITY-2892 warning from TestComplete

daypatrickb:
[SECURITY-2892](https://issues.jenkins.io/browse/SECURITY-2892): This was corrected by https://github.com/jenkinsci/testcomplete-plugin/pull/15 and was released in v2.9.2


Kevin-CB:
@daniel-beck , I was able to test the fix locally and I can confirm that there is no XSS anymore. 
At the same time, I wasn't able to identify any similar occurrences.

resources/warnings.json
```diff
@@ -15891,7 +15891,7 @@
     "versions": [
       {
         "lastVersion": "2.8.1",
```
daniel-beck, line 15893:
```suggestion
        "lastVersion": "2.9.1",
```

PR #733, opened 2023-08-09
Exclude 2.414.x from <2.416 warnings

daniel-beck:
https://github.com/jenkins-infra/update-center2/commit/f26a0a06fe7ba322e0e1259173a5622ec45e43cd#commitcomment-124002926

@NotMyFault Please do not comment directly on commits, as that means the committer is the single point of failure to notice and act on your comment (plus GH does not properly handle read/unread statuses).


NotMyFault:
> Please do not comment directly on commits

Given there was no PR and issues are disabled in this repository, what would you recommend for the future?

daniel-beck:
> what would you recommend for the future?

Infra helpdesk works.

PR #738, opened 2023-09-01
Decommission version 3.0.1 and 3.0.2 of Delphix plugin

ankit-patil-hubs:
We intend to remove Delphix v3.0.1 and v3.0.2 due to security vulnerabilities identified in them that were addressed with v3.0.3. We do not want any customers to download these versions and expose themselves to the risks.

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -877,3 +877,5 @@ delphix@2.0.1
 delphix@2.0.2
 delphix@2.0.3
 delphix@2.0.4
+delphix@3.0.1
+delphix@3.0.2


daniel-beck:
We're not generally supporting historical releases due to security issues. Jenkins admins will be informed through security warnings inside Jenkins if the version they have installed is unsafe.

ankit-patil-hubs:
@daniel-beck  From a Delphix customer and security standpoint, the recommendation is to remove these plugin versions. We see a risk of exposing internal credentials, which we do not want to take. Furthermore, the data shows that neither of these two plugin versions has been downloaded, giving us confidence to remove them at this early stage.

daniel-beck:
We're not removing old versions of plugins just because they have known vulnerabilities. The same applies to Jenkins (core) and all other plugins. We have no mechanism for this to scale, and there may be rational reasons for users to (temporarily) prefer older releases.

PR #746, opened 2023-10-21
Use public URLs for suspension reasoning

zbynek:
The INFRA- tickets were moved to GitHub, trying to access them leads to Jira login page.


lemeurherve (approved):
Thanks!

PR #748, opened 2023-10-30
Exclude 2.426.x from the 2023-10-18 core security warning

daniel-beck:
https://github.com/jenkins-infra/update-center2/commit/4200956686c55a93a78e25abac8409b5c739eb36#commitcomment-131192867


MarkEWaite (approved):
Thanks very much.  I tested and confirmed that 2.414 and 2.414.1 both show the security warning and that 2.414.2 and 2.414.3 do not show the security warning.  This follows a similar pattern as 2.414 by removing the "6" from one pattern inside the larger block and adding 2.426 to the concluding pattern.  I don't understand how that works in the interpretation of the regex for 2.414, but interactive testing shows that it works.  Since it works for 2.414, then it makes sense to use the same technique for 2.426.

PR #749, opened 2023-11-14
Mark SECURITY-3077 as fixed in 361.v44fea_4fc08d9

daniel-beck:
https://github.com/jenkinsci/hashicorp-vault-plugin/pull/314#issuecomment-1810168632


jglick (approved):
OK I think; technically https://github.com/jenkinsci/credentials-binding-plugin/pull/260 would have “fixed” it by sometimes breaking the build rather than leaking secrets.

daniel-beck:
> would have “fixed” it by sometimes breaking the build rather than leaking secrets

As long as that's not a dependency of the affected plugin, and you can use an old credentials-binding with hashicorp-vault-plugin 360, nothing changes wrt what the fix version is.

PR #753, opened 2023-11-21
[JENKINS-72353] Ban build-pipeline-plugin 2.0.0

MarkEWaite:
## Ban build-pipeline-plugin 2.0.0

[JENKINS-72353](https://issues.jenkins.io/browse/JENKINS-72353) reports that build pipeline plugin 2.0.0 generates stack traces into the Jenkins log because it incorrectly bundles the Jenkins test harness as a compile time dependency.

[PR 131](https://github.com/jenkinsci/build-pipeline-plugin/pull/131) fixes the dependency.

I'm not entirely sure that this is a severe enough error to ban the plugin version from distribution, especially since I am hopeful that @dalvizu will deliver a 2.0.1 release very quickly.

I'm proposing this ban in case there is a delay in delivering the 2.0.1 release or in case the repeated stack traces are severe enough to justify banning the plugin.

@daniel-beck can you offer guidance on this type of issue?  Should this be enough to ban the specific plugin release?

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -887,3 +887,8 @@ batch-task@1.16               # SECURITY-1025
 batch-task@1.17               # SECURITY-1025
 batch-task@1.18               # SECURITY-1025
 batch-task@1.19               # SECURITY-1025
+
+# Remove build-pipeline-plugin that incorrectly bundles Jenkins test harness
+# https://issues.jenkins.io/browse/JENKINS-72353 for the bug report
+# https://github.com/jenkinsci/build-pipeline-plugin/pull/131 for the fix
+build-pipeline-plugin@2.0.0


MarkEWaite:
I'm also quite happy if this pull request is closed because the repeated stack traces are not disruptive enough to justify banning the specific release of the plugin.

MarkEWaite:
Closing this because 2.0.1 has been released with the fix.

daniel-beck:
> generates stack traces into the Jenkins log

IMO this doesn't look relevant enough to suspend distribution. Besides the fact that this looks like an easy to ignore logger (set it to log at `SEVERE`), it seems limited to startup/component loading, rather than continuously all the time.

MarkEWaite:
> > generates stack traces into the Jenkins log
> 
> IMO this doesn't look relevant enough to suspend distribution. Besides the fact that this looks like an easy to ignore logger (set it to log at `SEVERE`), it seems limited to startup/component loading, rather than continuously all the time.

Thanks very much.  I hadn't done enough analysis to identify how widely it might be visible.

PR #761, opened 2024-01-24
Exclude 2.440.* from warnings for 2024-01-24 security advisory

Kevin-CB:
I'm not 100% confident with my regex change, I'm asking for a review from @daniel-beck


daniel-beck (approved):
Same general approach as https://github.com/jenkins-infra/update-center2/blob/5862a79d2bd9da4356b9a732b7a2cc8b62370f52/resources/warnings.json#L16613, so should be fine.

PR #654, opened 2022-10-29
Mark plugins without maintainers as up for adoption

daniel-beck:
Implement https://github.com/jenkins-infra/repository-permissions-updater/pull/2891#issuecomment-1290221864

Waiting for https://github.com/jenkins-infra/repository-permissions-updater/issues/2911 to proceed.

PR #762, opened 2024-02-15
Marks Chaos Butler plugin as up or adoption

alecharp:
I propose to mark the Chaos Butler plugin as up for adoption, because:
- no developer is listed in https://github.com/jenkins-infra/repository-permissions-updater/blob/8253c2e335a6479630dbfe679b2ee06e63a09fdc/permissions/plugin-chaos-butler.yml#L8
- there was one release done 7years ago with no additional work
- there is no activity on the open pull requests

cc @stephenc as last user active on the repository.


daniel-beck:
Please set the topic on the repository instead, that scales a lot better than this centrally controlled metadata file 😉

alecharp:
I would agree but I don't have access to the repository.
Be to fair, this is also because we are still not automatically marking plugin with no developers as abandoned. 
The jenkins.io page about adopting a plugin mention this as an alternative..

There is more than 300 plugins with no developers referenced, I'd rather not open that many pull request.
Maybe this is a discussion to have on the mailing list.

(Opened this one to focus attention on the situation)

daniel-beck:
> because we are still not automatically marking plugin with no developers as abandoned

Fix https://github.com/jenkins-infra/repository-permissions-updater/issues/2911 and we can do that 😉

daniel-beck:
This is awkward:

> I would agree but I don't have access to the repository.

https://github.com/jenkinsci/chaos-butler-plugin/ has the label, but due to https://github.com/jenkinsci/chaos-butler-plugin/commit/845d6b0b0b27ba2a6e02afca972f2b520e9b5993 being unreleased, it's ineffective.

alecharp:
I guess this is another good reason to have the label added directly on the update-center, no?

daniel-beck:
Might be a case for https://github.com/jenkins-infra/update-center2/blob/master/resources/wiki-overrides.properties to also fix the broken GH link from https://plugins.jenkins.io/chaos-butler/

Alternatively I could try to rename the GH repo back and forth to create a redirect. 🤔

alecharp:
It's odd that the redirect is not already in place. I don't know the history of the renaming, was it from the hosting already?

alecharp:
Thanks to #654, this is now irrelevant.

PR #765, opened 2024-03-01
Delete/Ignore version 24.1 of the hpaat plugin

andreibangau99:
Hello!

We accidentally released the [24.1 version](https://repo.jenkins-ci.org/ui/repos/tree/General/releases/org/jenkins-ci/plugins/hp-application-automation-tools-plugin/24.1/hp-application-automation-tools-plugin-24.1.jar?clearFilter=true) of the [plugin](https://plugins.jenkins.io/hp-application-automation-tools-plugin) too early, while trying to do a dry run.

We would like this version to be removed or at least not to be available for use.
On the 12th of march we will need to release the same version of the plugin -- 24.1. Can this be done, since here it will be ignored? Or we will have to make the release with a new version?

Thank you!


daniel-beck:
>  On the 12th of march we will need to release the same version of the plugin -- 24.1. Can this be done, since here it will be ignored? Or we will have to make the release with a new version?

The version number is now used by the suspended release. If you do not want to restore publication of this release by reverting this PR (e.g., when nothing changes anyway between now and then) you need to use a different version number. Since admins may have already installed 24.1, I recommend a greater version number than that.

andreibangau99:
Thank you for the clarification!

PR #767, opened 2024-03-18
Resolving SECURITY-2979

michael-doubez:
Solved in version
See 3.0 https://github.com/jenkinsci/oic-auth-plugin/releases/tag/oic-auth-3.0


Kevin-CB (approved):
I tested the fix, I confirm it's fixed

PR #775, opened 2024-04-13
Suspend obsolete plugins

zbynek:
* SLAdiator: Website of the service is dead, [last website snapshot](https://web.archive.org/web/20170331010759/https://sladiator.com/) 2017 [last tweet](https://twitter.com/sladiator) in 2013
* Flexteam: has no functionality, consists of [only one empty source file](https://github.com/jenkinsci/flexteam-plugin/blob/master/src/main/java/org/jvnet/hudson/plugins/flexteam/SidebarAction.java)
* Delta Cloud -- service retired in 2015 https://deltacloud.apache.org/


MarkEWaite (approved):
Thanks for detecting those obsolete services.

PR #785, opened 2024-05-29
Update to bouncycastle jdk18on with version 1.78.1

pathob:
The Jenkins update-center2 was still using a bouncycastle version from September 2008. Even though I exactly followed the instructions from the update-center2's README.adoc file, I was not able to generate an OpenSSL key (and certificate) that would work with that bouncycastle version. Debugging into the library revealed that the old bouncycastle version doesn't support reading private keys starting with the line `-----BEGIN PRIVATE KEY-----`.

Instead, bouncycastle only supported one of the following starting lines for private keys:

- `-----BEGIN RSA PRIVATE KEY-----`
- `-----BEGIN DSA PRIVATE KEY-----`
- `-----BEGIN EC PRIVATE KEY-----`

In the case of a starting line of `BEGIN PRIVATE KEY`, the `readObject()` method of the `PEMReader` would just return `null` and cause the update-center2 to crash.

However, the documented command `openssl genrsa -out demo.key 4096` also generates a key file starting with `BEGIN PRIVATE KEY` when using modern versions of OpenSSL (version 3.0.13 in my case).

For this reason, this change updates the used bouncycastle library to the most recent version, with which I was able to sign the files again.


pathob:
Hi @daniel-beck , it would be great if you would consider accepting this MR, since this is the missing puzzle piece for generating a corporate update center for my client.

Unfortunately, the included bouncycastle dependencies are not available in https://repo.jenkins-ci.org/public so that I included Maven central for developing and testing purposes. If you could arrange for the library to be available in the Jenkins repository, I would remove Maven Central again.

daniel-beck:
@pathob Apologies for the delayed response here, I missed the notification. Thanks for this PR, I will try to review that in the next few weeks.

Meanwhile I think we encountered the same problem when we last created the new yearly cert, and the solution (per Gitter notes) was to pass the `-traditional` flag to `openssl genrsa`. Perhaps that's a suitable workaround for you if you don't want to deal with a fork.

pathob:
@daniel-beck Thanks for the info about the `-traditional` flag. It would be great if you would find time to review this MR, otherwise I think it should be a good idea to add it to the README, because debugging this was quite frustrating. Thank you!

pom.xml
```diff
@@ -269,6 +274,10 @@
   </dependencies>
 
   <repositories>
+    <repository>
+      <id>central</id>
+      <url>https://repo1.maven.org/maven2/</url>
+    </repository>
```
daniel-beck, line 280: Maven central is defined in the super pom. Could you check whether this is actually needed?

daniel-beck:
@pathob PTAL at the recent changes, do you see any potential problems?

PR #786, opened 2024-05-30
Update properties for ecutest and ecu-test-execution plguin

TT-S3-D1:
* added deprecated for https://github.com/jenkinsci/ecutest-plugin
* added properties for https://github.com/jenkinsci/ecu-test-execution-plugin


MartinGroscheTT (approved):

daniel-beck:
We prefer that plugin labels be applied through GitHub topics as documented at https://github.com/jenkins-infra/update-center2?tab=readme-ov-file#categorizing-plugins

This resource file remains supported for cases where that's not possible. If that's the case, please explain why.

TT-S3-D1:
Hi @daniel-beck 
we have done both to be safe. If this is not necessary, you can reject this change.

daniel-beck:
> Do you mean only the topic for the deprecation of the old plugin or is this valid for all labels / topics?

All supported topics, see https://github.com/jenkins-infra/update-center2?tab=readme-ov-file#github

PR #732, opened 2023-08-08
Suspend Azure Dev Spaces plugin, service discontinued

MarkEWaite:
## Suspend Azure Dev Spaces plugin, service discontinued

[Azure updates June 14, 2021](https://azure.microsoft.com/en-us/updates/azure-dev-spaces-retired-on-may-15-2021/) announces the retirement of Azure Dev Spaces May 15, 2021.

The [Microsoft learn site](https://learn.microsoft.com/en-us/previous-versions/azure/dev-spaces/) points to "Bridge to Kubernetes" as the replacement.

[JENKINS-71776](https://issues.jenkins.io/browse/JENKINS-71776) asks to suspend azure-dev-spaces becuse the service is no longer available.

PR #587, opened 2022-04-22
Suspend plugins for discontinued services

zbynek:
* [Fabric Beta was replaced by Firebase](https://github.com/jenkinsci/fabric-beta-publisher-plugin#deprecation-notice) -- users should migrate to Firebase
* Kmap: this plugin depends on Keivox Kmap service previously available at keivox.com (domain for sale), company's online activity seems to end in [2015](https://twitter.com/keivoxdev)
* Discobit AutoConfig: domain discobit.com is up for sale

PR #791, opened 2024-08-09
Suspend windows-slaves and multi-slave-config plugins

MarkEWaite:
## Suspend windows-slaves and multi-slave-config plugins

We suspend distribution of plugins that depend on a service that no longer exists or is end of life like:

* https://github.com/jenkins-infra/update-center2/pull/732
* https://github.com/jenkins-infra/update-center2/pull/708
* https://github.com/jenkins-infra/update-center2/pull/676
* https://github.com/jenkins-infra/update-center2/pull/587

This is a similar case to services that are end of life.  Windows is not end of life, but the services required for this plugin have been end of life since at least March 2023. The plugin was deprecated as noted in [README](https://github.com/jenkinsci/windows-slaves-plugin?tab=readme-ov-file#notice-of-deprecation) almost two years ago in [November 2022](https://github.com/jenkinsci/windows-slaves-plugin/commit/cdafd13807267b535b13bbb25d9cc72376aa4594).

The [plugin installation statistics](https://stats.jenkins.io/pluginversions/windows-slaves.html) show that there are still almost 80,000 installations of the plugin, but only 22% of those installations are on 2.426.3 or newer.  Most installations of the plugin are not upgrading their Jenkins controller version.

Users with the plugin already installed will not be affected by distribution being suspended.  Users with configuration as code definitions that include the plugin will need to remove it from their configuration as code definition.  They will receive an error message when they attempt to download the plugin version.

Implied dependencies on the plugin were removed from [Jenkins 2.386](https://www.jenkins.io/changelog/2.386/) (10 Jan 2023)

* https://github.com/jenkinsci/jenkins/pull/7568

The plugin was removed from the setup wizard in [Jenkins 2.379](https://www.jenkins.io/changelog/2.379/) (22 Nov 2022)

* https://github.com/jenkinsci/jenkins/pull/7414

Fixes https://github.com/jenkins-infra/helpdesk/issues/4221

The multi-slave-config plugin has additional reasons why its distribution should be suspended.  It uses Prototype.js and so is not usable with current Jenkins releases.  It uses the YahooUI JavaScript library and so will not be usable with upcoming Jenkins versions that will remove the YahooUI JavaScript library.  It does not have an active maintainer and was last released 10 years ago.  It has [less than 1000 installations](https://stats.jenkins.io/pluginversions/multi-slave-config-plugin.html).

### DIFF
--- a/resources/artifact-ignores.properties
+++ b/resources/artifact-ignores.properties
@@ -164,6 +164,9 @@ mber                             # service discontinued -- https://wiki.jenkins.
 # superseded by Relution Publisher
 mcap-eas-plugin = https://github.com/mwaylabs/jenkins-mcap-eas-plugin
 mogotest                         # deprecated: mogotest service no longer exists
+# Windows DCOM as used by this plugin is blocked by Microsoft security from CVE-2021-26414
+# Depends on the windows-slaves plugin that is also suspended
+multi-slave-config = https://github.com/jenkins-infra/helpdesk/issues/4221#issuecomment-2277844595
 # service discontinued
 mypeople = https://wiki.jenkins-ci.org/display/JENKINS/MyPeople+Plugin
 # service discontinued
@@ -258,6 +261,8 @@ vagrant-plugin                    # renamed to vagrant, herpderp
 vessel = https://groups.google.com/d/msg/jenkinsci-dev/L34eAMMWA5o/-AxtRdGsAAAJ
 weibo4jenkins                    # renamed to weibo
 whitesource = https://docs.mend.io/bundle/unified_agent/page/jenkins_plugin.html
+# Windows DCOM as used by this plugin is blocked by Microsoft security from CVE-2021-26414
+windows-slaves = https://github.com/jenkinsci/windows-slaves-plugin?tab=readme-ov-file#notice-of-deprecation
 wsnotifier                       # renamed to websocket
 # renamed to xltestview-plugin
 xltest-plugin = https://github.com/jenkinsci/xltestview-plugin/commit/ea034f9929b520e63b9ce15aed9bdb62354146cf


MarkEWaite:
The "down side" of suspending the plugins is there will be users that will raise questions when they are no longer able to download the windows-slaves plugin that is listed in their plugins.txt file.  The number of installations of the windows-slaves is steadily decreasing.  It is down from 106,000 12 months ago to 80,000 now.

daniel-beck:
>  Windows is not end of life, but the services required for this plugin have been end of life since at least March 2023.

What does the "deprecation" registry key and date etc. mean for Windows Server 2012 and other outdated platforms people are still using?

MarkEWaite:
> > Windows is not end of life, but the services required for this plugin have been end of life since at least March 2023.
> 
> What does the "deprecation" registry key and date etc. mean for Windows Server 2012 and other outdated platforms people are still using?

That's a very good question.  The [Microsoft site](https://learn.microsoft.com/en-us/lifecycle/announcements/windows-server-2012-r2-end-of-support) says that standard support for Windows Server 2012 and Windows Server 2012 R2 ended in October 2023.  However, they offer extended security support through Oct 2026 if purchased or if Windows Server 2012 computer is hosted on Azure.  Windows Server 2012 is an interesting case because it is a sort of hybrid between "no more security updates unless you purchase them" (like RHEL 7) and "security updates are available at no charge".  The site says:

> If you cannot upgrade to the next version, you will need to use Extended Security Updates (ESUs) for up to three years. ESUs are available for free in Azure or need to be purchased for on-premises deployments. 

I wasn't aware of that hybrid model offered by Microsoft, though I believe a similar model is offered by Google Cloud and other cloud providers for support of Red Hat Enterprise Linux 7 and Ubuntu 18.04.  Extended support can be purchased for Google Cloud instances of Red Hat Enterprise Linux 7 and Ubuntu 18.04.  Other cloud providers offer the same purchase option for users of those operating systems that are outside of standard support.  In the case of Windows Server 2012 and Windows Server 2012 R2, it appears that only Azure users receive the extended support without extra payment.

I think that the Windows Server 2012 Extended Security Updates (ESUs) is similar enough to [Red Hat Enterprise Linux 7 Extended Lifecycle Support](https://www.redhat.com/en/blog/announcing-4-years-extended-life-cycle-support-els-red-hat-enterprise-linux-7) and [Ubuntu 18.04 Extended Security Maintenance](https://fridge.ubuntu.com/2023/06/17/extended-security-maintenance-for-ubuntu-18-04-lts-began-on-may-31-2023/) that we can treat them similarly, as though they are not supported by the vendor unless an additional support contract is purchased.  In the case of Red Hat Linux 7 and Ubuntu 18.04, we ended support from Jenkins when the standard support period ended because we were unwilling to continue testing an operating system that would require us to purchase an extended support contract.

However, I'm open to the argument that this is premature and should wait until October 2026 and the end of extended security updates for Windows Server 2012.

MarkEWaite:
@daniel-beck does my [further explanation](https://github.com/jenkins-infra/update-center2/pull/791#issuecomment-2280976227) clarify why I think the two plugins should be suspended?

Do you have a preference for our handling of these two plugins?

If your preference is to not suspend these plugins at this time, I'll close this pull request and close the help desk ticket that started the process.

daniel-beck:
Thanks for the ping.

Windows Server 2012 was a random old release I would imagine still supports this method, so how long that specifically is supported seems secondary. Sorry if my comment was misleading there.

That said, it looks like the KB linked from WMI Agents docs answers this, it seems to apply all recent and not-so-recent releases of Windows (which seems crazy unlikely to me, but I'm too removed from the MS ecosystem to evaluate this), since Windows 7 and Server 2008.

I will give this another look tomorrow and then merge.

PR #795, opened 2024-09-04
Deprecate plugins incompatible with EE9/Spring Security 6 update

daniel-beck:
These plugins appear incompatible with the changes in https://github.com/jenkinsci/jenkins/pull/9672 that are part of Jenkins 2.475+.

* https://plugins.jenkins.io/read-only-configurations/ has last been released 10 years ago and looks like it's unnecessary since [JEP-224](https://github.com/jenkinsci/jep/tree/master/jep/224).
* https://plugins.jenkins.io/jobgenerator/ has last been released 10 years ago and has had an unresolved vulnerability for more than two years.

Therefore I'm proposing their deprecation.

CC @basil


basil (approved):
Thanks!

PR #796, opened 2024-09-05
Undeprecate Maven Repository Scheduled Cleanup

cpfeiffer:
This is the inverse of 8798cb525a8a20a405dfd208a4fd7fa4a25e540f

maven-repo-cleaner has been released and works just fine.

See #603.


daniel-beck:
Doesn't look like @olamy intends to actually maintain the plugin, but I guess the original reason to deprecate is now obsolete (perhaps replaced by a weaker one).

cpfeiffer:
Probably not the correct place to propose this, but if needed, I would volunteer to maintain the plugin.

daniel-beck:
Right, not a blocker for this PR, just generally -- unmaintained plugins aren't great :)

PR #798, opened 2024-09-11
hotfix(publish) avoid publication failures by using the VM specific hostname

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2344340606

The new UC system (in azure) uses the ZIP credential with proper hostname/etc. for rsync (and others).
But the script still specifies (and need) to update the current UC system (in AWS).

### DIFF
--- a/site/publish.sh
+++ b/site/publish.sh
@@ -30,7 +30,7 @@ fi
 
 if [[ "${run_stages[*]}" =~ 'sync-plugins' ]]
 then
-    UPDATES_SITE="updates.jenkins.io"
+    UPDATES_SITE="aws.updates.jenkins.io"
     RSYNC_USER="mirrorbrain"
 
     ## $download_dir folder processing


site/publish.sh
```diff
@@ -30,7 +30,7 @@ fi
 
 if [[ "${run_stages[*]}" =~ 'sync-plugins' ]]
 then
-    UPDATES_SITE="updates.jenkins.io"
+    UPDATES_SITE="aws.updates.jenkins.io"
```
MarkEWaite, line 33: Is this pointing to the virtual machine on AWS or should it be azure.updates.jenkins.io ?
```suggestion
    UPDATES_SITE="azure.updates.jenkins.io"
```

dduportal: It's expected to point to the AWS VM (aka. `pkg.origin.jenkins.io`) to keep it updated.

I've updated the PR body with explanation and link to the issue: does it explain better?

dduportal: @MarkEWaite many thanks for double checking! I'm self-merging as the checks are OK (and to control the update_center2 generation)

PR #799, opened 2024-09-23
chore(publish) add support of the http (unsecured) Update Center

dduportal:
(untested yet: WiP)


dduportal:
- Tested with success on production (with real life data) (see build 935633).
- This PR does NOT change the generation of metadatas, only the publication.

As such, I'm merging it to allow the infra team to proceed

PR #802, opened 2024-09-25
Suspend `shared-library-version-override`

daniel-beck:
https://github.com/jenkinsci/shared-library-version-override-plugin/blob/786074c9fce7adeaefcdf6b869304be19464c07b/src/main/java/io/jenkins/plugins/shared_library_version_override/FolderConfigurations.java#L78-L80 is unsafe. Since this plugin was first released only today, I propose we're suspend this plugin. Once this issue is resolved, we can restore distribution.

CC @c3p0-maif as maintainer


MarkEWaite (approved):

jonesbusy:
Is it fixed by https://github.com/jenkinsci/shared-library-version-override-plugin/pull/7 ?

c3p0-maif:
Hi @jonesbusy, the last release is still on testing.

PR #804, opened 2024-09-26
Tag the Gatling plugin with adopt-this-plugin

slandelle:
Hello,

We of Gatling no longer maintain this plugin (we have a distinct one for Gatling Enterprise) and would like to mark it for adoption.

Thanks!


zbynek:
To set a label for a plugin you can just add a topic to the GitHub repository -- it's preferred over using the label file here.
https://github.com/jenkins-infra/update-center2/pull/762#issuecomment-1945949807

slandelle:
@zbynek I did both as instructed in the official documentation: https://www.jenkins.io/doc/developer/plugin-governance/adopt-a-plugin/#how-can-i-mark-a-plugin-for-adoption.

zbynek:
The documentation mentions two ways to do the same thing, there's no benefit in doing both.

The plugin is displayed as "up for adoption" on the plugin site (https://plugins.jenkins.io/gatling/) so I think this PR can be closed.

slandelle:
@zbynek My point was that you might want to update your documentation if the second way is not recommended and the first one suffices. Anyway, closing as instructed.

PR #809, opened 2024-10-11
Deprecate and suspend `synopsys-sigma` plugin

sboyd-synopsys:
Being replaced with a rebranded version that is now hosted at https://github.com/jenkinsci/black-duck-sigma-plugin


sboyd-synopsys:
The https://github.com/jenkinsci/black-duck-sigma-plugin plugin has officially been released now, so this plugin is good to be deprecated/suspended.

PR #810, opened 2024-10-17
Use absolute URLs for links from download indexes

daniel-beck:
Fixes https://github.com/jenkins-infra/helpdesk/issues/4311

PR #812, opened 2024-10-18
fix(site/htaccess) only use Apache `mod_rewrite` by replacing `RedirectMatch` by `RewriteRule` directives

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2380569628 and https://github.com/jenkins-infra/helpdesk/issues/4311

After a discussion with @daniel-beck , we realized that some of the issues on the new update center could be solved using smart request distribution on Apache. 

The mix of `mod_rewrite` (`RewriteRule`) and `mod_alias` (`RedirectMatch` ) in the generated `htaccess` files makes the rules ordering hard to determine as Apache evaluates all rewrite rules and apply the module's own internal priority, and then evaluate the redirect matchings.

We were forced to use `RedirectMatch` in https://github.com/jenkins-infra/update-center2/pull/790 to ensure the new UC fallback to `mirrors.updates.jenkins.io` was working as expected without breaking existing redirections.

Using rewrite rules allows more powerful setup such as using internal variables, applying conditions, etc.

This PR:

- Add test case to cover all current `RedirectMatch` rules (and the test works obviously)
- Replaces the `RedirectMatch` rules by `RewriteRule` directives (with test still passing)
- Changing the new UC fallback rule to a `RewriteRule` which tests the absence of the file specified by the request in order to send to the mirrors. As such, if we copy any file (such as `uctest.json` or `/download/**/*.html`) in the Apache document roots, then we can control which files are served from Azure or from the mirrors.

PR #813, opened 2024-10-21
chore(wrapper/publish) serve `uctest.json` and the HTML in the virtual tree `download/` from Apache instead of mirrors

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4311#issuecomment-2419583755 (second bullet).

~Closes 810 (as not needed anymore)~

This PR is a follow up of #812  for the new [Update Center](https://github.com/jenkins-infra/helpdesk/issues/2649) using Apache2 + Mirrorbits.

It updates the publication script to serve the following file from Apache instead of mirrors:

- `uctest.json` => solves the HTTP/404 on azure.updates.jenkins.io/whatever?uctest
- `download/**/*.html` (only these files are present: other are sent to get.jenkins.io` => should solve the broken links in these HTML pages

As #812 did add a condition on "the file must not exist in httpd documentroot" to redirect incoming requests to the UC mirrors, Apache will serve these files.

As discussed with @daniel-beck:

- These files will count in the (billed) outbound bandwidth as served by Apache, but the amount of traffic is really low so it does not even count
- These files won't benefit from the new "mirror proximity" feature. But we do not really care as it is a really low usage
- We won't need to define an absolute URL (constraining us for years to a scheme and an hostname: not really flexible)

### DIFF
--- a/site/publish.sh
+++ b/site/publish.sh
@@ -149,6 +149,8 @@ then
         --prune-empty-dirs `# Do not copy empty directories` \
         --exclude='updates/' `# Exclude ALL 'updates' directories, not only the root /updates (because symlink dereferencing create additional directories` \
         --exclude='.htaccess' `# Exclude every .htaccess files` \
+        --exclude="uctest.json" `# Service Applicative Healthcheck (empty JSON)` \
+        --exclude="download/***" `# Virtual Tree of the download service, redirected to get.jio, with only HTML version listings with relative links to UC itself` \
         "${www2_dir}"/ ./www-content/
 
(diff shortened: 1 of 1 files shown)


daniel-beck:
> Closes https://github.com/jenkins-infra/update-center2/pull/810 (as not needed anymore)

Sure, but as we've discussed, it's a good idea anyway to link to the canonical URLs.

dduportal:
> > Closes #810 (as not needed anymore)
> 
> Sure, but as we've discussed, it's a good idea anyway to link to the canonical URLs.

Updated the body then, to avoid PR being closed

dduportal:
Merging as it's only scoped to the `publish.sh` script to solve the new UC remaining errors

PR #814, opened 2024-10-23
Add CODEOWNERS for warnings metadata and the publishing script

daniel-beck:
@dduportal WDYT?

CC @Kevin-CB.


MarkEWaite (approved):
I think that makes sense

Kevin-CB (approved):

dduportal (approved):
Excellent idea, thanks!

PR #818, opened 2024-11-03

itaibh:
Hello!

I accidentally released version 1.16.4 of applitools-eyes, which doesn't work.

I'd like to block this specific version from being downloaded. In addition I'll re-release the last working version with a new version number.

Thanks.


daniel-beck:
Given the successor already exists, this is not needed.

Please also note that this expects the artifact ID, i.e., `applitools-eyes`, so this PR wouldn't have worked anyway.

PR #819, opened 2024-11-06
chore(wrappers/publish) ensure Apache has access to symlinked folders

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4375

This PR aims at fixing the issue:

- `azcopy sync` does not provide any symlink option
- We cannot use `azcopy copy` (which has symlinks options)

So we take advantage of `rsync` to do the "symlink dereferencing" for us.

Note: the impact in timing can be ignored as this task is in parallel of S3 which takes WAY more time


smerle33 (approved):
🚀

PR #824, opened 2024-11-06
chore(wrappers/publish) ensure Apache has access to symlinked folders

smerle33:
Related to https://github.com/jenkins-infra/helpdesk/issues/4375

This PR aims at fixing the issue:

    azcopy sync does not provide any symlink option
    We cannot use azcopy copy (which has symlinks options)

So we take advantage of rsync to do the "symlink dereferencing" for us.

Note: the impact in timing can be ignored as this task is in parallel of S3 which takes WAY more time

lets add a mkdir to avoid 

```
symlink has no referent: "/home/jenkins/agent-workspace/workspace/update_center/www2/current/updates"
```

### DIFF
--- a/site/publish.sh
+++ b/site/publish.sh
@@ -159,6 +159,8 @@ then
         --exclude='*' `# Exclude all other files` \
         "${www2_dir}"/ "${content_dir}"/
 
+    # create an empty folder for the cawler content to avoid error with deference symlinks
+    mkdir -p "${www2_dir}"/current/updates
     # Prepare www-redirections-*secured/ directories, same content as $www2_dir (to allow directory listing) but with dereferenced symlinks, dedicated to httpd services
     rsync --archive --verbose \
         --copy-links `# derefence symlinks` \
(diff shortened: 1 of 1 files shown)


site/publish.sh
```diff
@@ -159,23 +159,28 @@ then
         --exclude='*' `# Exclude all other files` \
         "${www2_dir}"/ "${content_dir}"/
 
-    # Prepare www-redirections-*secured/ directories, same content as $www2_dir (to allow directory listing), dedicated to httpd services
-    cp -r "${www2_dir}" "${httpd_secured_dir}"
-    cp -r "${www2_dir}" "${httpd_unsecured_dir}"
+    # create an empty folder for the cawler content to avoid error with deference symlinks
+    mkdir -p "${www2_dir}"/current/updates
```
dduportal, line 163:
```suggestion
    mkdir -p "${www2_dir}"/updates
```

dduportal (approved):
LGTM!

Tested  the `rsync` part during a pair review

PR #827, opened 2024-11-27
chore(wrappers/publish) add 3 new rsync targets on the new data service

dduportal:
This PR is a retry of #825 which was reverted by #826 due to rsync errors on the service

We fixed (and tested in details) the new rsync now it's been using NFS instead of SMB for the shared volume: https://github.com/jenkins-infra/helpdesk/issues/4402


dduportal (commented):
Let's retry (advisory is out): I'm watching the result or reverting

PR #829, opened 2024-11-28
Revert "Revert "chore(wrappers/publish) add 3 new rsync targets on the new data service""

dduportal:
Reverts jenkins-infra/update-center2#828

As soon as https://github.com/jenkins-infra/kubernetes-management/pull/5959 is deployed, we can safely start using rsync tasks to fill the NFS shared volume


dduportal (commented):
Let's roll: the SSH server fingerprint is present for the user `jenkins`

PR #831, opened 2024-12-04
feat(publish): add archives.jenkins.io in the rsync process for update-center

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4401#issuecomment-2515106384
add archives.jenkins.io to the parallel steps for rsync in order to add it as a failback mirror

depends on https://github.com/jenkins-infra/charts-secrets/pull/34


dduportal (approved):
Crawler worked well with the new credential (see https://github.com/jenkins-infra/charts-secrets/pull/34#issuecomment-2532159698)

Let's roll for initial phase

PR #636, opened 2022-08-12
Add 'releaseTimestamp' with second resolution to plugin-versions.json

daniel-beck:
Add higher resolution field for release timestamps for https://github.com/jenkins-infra/plugin-site/issues/1220.

The field name and date format is consistent with the corresponding field in `update-center.json`

As a side effect, this adds 45 bytes for each plugin release (of which there are 26k), roughly increasing file size by 9%.

PR #835, opened 2025-01-24
Generate `timestamp.txt`

daniel-beck:
`Last-Modified` headers of the metadata files are no longer a reliable indicator of age, as these are different between the different hosts that respond to those requests since recent infra changes (archives/fallback vs. CF).

Downloading all of `update-center.actual.json` to get the timestamp from https://github.com/jenkins-infra/update-center2/pull/636 is impractical, and wasteful if done repeatedly in a short amount of time.

So this PR adds the new file `timestamp.txt` that has the same content as the value of `.generationTimestamp` in `update-center.actual.json` would be.

### DIFF
--- a/site/generate.sh
+++ b/site/generate.sh
@@ -112,12 +112,12 @@ function sanity-check {
 # This supports updating Jenkins (core) once a year while getting offered compatible plugin updates.
 for version in "${WEEKLY_RELEASES[@]}" ; do
   # For mainline, advertising the latest core
-  generate --limit-plugin-core-dependency "$version" --write-latest-core --latest-links-directory "$WWW_ROOT_DIR/dynamic-$version/latest" --www-dir "$WWW_ROOT_DIR/dynamic-$version"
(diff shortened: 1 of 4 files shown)


MarkEWaite (approved):

dduportal (approved):
(FTR: caused by https://github.com/jenkins-infra/helpdesk/issues/2649).

LGTM!

site/generate.sh
```diff
@@ -112,12 +112,12 @@ function sanity-check {
 # This supports updating Jenkins (core) once a year while getting offered compatible plugin updates.
 for version in "${WEEKLY_RELEASES[@]}" ; do
   # For mainline, advertising the latest core
-  generate --limit-plugin-core-dependency "$version" --write-latest-core --latest-links-directory "$WWW_ROOT_DIR/dynamic-$version/latest" --www-dir "$WWW_ROOT_DIR/dynamic-$version"
+  generate --limit-plugin-core-dependency "$version" --write-latest-core --write-timestamp --latest-links-directory "$WWW_ROOT_DIR/dynamic-$version/latest" --www-dir "$WWW_ROOT_DIR/dynamic-$version"
```
daniel-beck, line 115: FTR the changes in this file were stupid as it's still set up to download and run 3.16 without the Java changes.

PR #840, opened 2025-02-03
Incorporate plugin health score in JSON metadata

alecharp:
This adds the plugin health, from the https://github.com/jenkins-infra/plugin-health-scoring project, into the update-center output.

This is the first step to be able to display the scores in the Plugin Manager.

Corresponding core PR: https://github.com/jenkinsci/jenkins/pull/10351

# Testing

I tested on a limited number of plugins and it worked. I'm testing now with all plugins.


src/main/java/io/jenkins/update_center/HealthScores.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import com.alibaba.fastjson.JSON;
+import io.jenkins.update_center.util.HttpHelper;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+
+public class HealthScores {
+    private static final String HEALTH_SCORES_URL = "https://reports.jenkins.io/plugin-health-scoring/scores.json";
```
alecharp, line 14: Rather than using the application endpoint, which generate the output on demand, this stores a static file, generated every hour IIRC.

dduportal (commented):
First set of feedback/point of attention:

- What is the impact on the size of the JSON metadatas? I have vague recollections of impacts due to adding even an integer field causing challenges. Of course it might be hard to evaluate: if it is the case, I would propose to start with generating both JSONs files (with and without the PH score), we keep deploying the "without PH score" and measure the impact in size and build/deploy time before going forward.

- This PR adds a dependency on the service reports.jenkins.io. If the service is down or not reachable, what would be the expected fall backs? This webservice is "HA" (e.g. it has a QoS bound to the underlying Azure Kubernetes cluster today) but could be down.

alecharp:
> * What is the impact on the size of the JSON metadatas? I have vague recollections of impacts due to adding even an integer field causing challenges. Of course it might be hard to evaluate: if it is the case, I would propose to start with generating both JSONs files (with and without the PH score), we keep deploying the "without PH score" and measure the impact in size and build/deploy time before going forward.

I'll generate both and let you down the time required for the generation and the size the the files in each case.

> * This PR adds a dependency on the service reports.jenkins.io. If the service is down or not reachable, what would be the expected fall backs? This webservice is "HA" (e.g. it has a QoS bound to the underlying Azure Kubernetes cluster today) but could be down.

Technically, we already have this dependency: 
https://github.com/jenkins-infra/update-center2/blob/master/src/main/java/io/jenkins/update_center/MaintainersSource.java#L47-L48. However, I'll try to see what happens and improve the code if needed for this situation.

dduportal:
> I'll generate both and let you down the time required for the generation and the size the the files in each case.

I wasn't sure it was possible or easy for you (as the production dataset could be heavy, not even speaking about the cache...). Thanks!

> Technically, we already have this dependency: https://github.com/jenkins-infra/update-center2/blob/master/src/main/java/io/jenkins/update_center/MaintainersSource.java#L47-L48. However, I'll try to see what happens and improve the code if needed for this situation.


OH: I was not aware. Then your change here does NOT add any dependency but reuse existing. It makes sense then (as per my infra hat: I can answer for the scope of the code itself).

alecharp:
So for the size:

```bash
❯ ls -al with*
with-health-score:
total 18528
drwxr-xr-x   5 alecharpe  *      160  3 fév 15:25 .
drwxr-xr-x  20 alecharpe  *      640  3 fév 15:28 ..
-rw-r--r--   1 alecharpe  *  3158562  3 fév 15:25 update-center.actual.json
-rw-r--r--   1 alecharpe  *  3158584  3 fév 15:25 update-center.json
-rw-r--r--   1 alecharpe  *  3158787  3 fév 15:25 update-center.json.html

without-health-score:
total 18312
drwxr-xr-x   5 alecharpe  *      160  3 fév 15:28 .
drwxr-xr-x  20 alecharpe  *      640  3 fév 15:28 ..
-rw-r--r--   1 alecharpe  *  3124324  3 fév 15:28 update-center.actual.json
-rw-r--r--   1 alecharpe  *  3124346  3 fév 15:28 update-center.json
-rw-r--r--   1 alecharpe  *  3124549  3 fév 15:28 update-center.json.html
```

For the timing, I ran the commands with an existing cache as downloading the files should not be part of the problem. As the health report is accessed / loaded once, this changeset should not be the biggest source of issues. Compared to accessing all the plugin GitHub repositories for example.

* without the health score: `java -Dfile.encoding=UTF-8 -jar  --id default --www-dir without-health-score  14,61s user 1,77s system 14% cpu 1:55,94 total`
* with the health score: `java -Dfile.encoding=UTF-8 -jar  --id default --www-dir with-health-score  13,82s user 1,71s system 12% cpu 2:02,35 total`

I'm still looking at what happens with no access to reports service.

alecharp:
So apparently, with reports.jenkins.io is down (blocked it locally), the `master` branch is not handling that well: it breaks the generation entirely. 
I don't think there is a point of testing this scenario for this changeset. But that's something we can track to improve in the future.

src/main/java/io/jenkins/update_center/HealthScores.java
daniel-beck: True per https://github.com/jenkins-infra/infra-reports/blob/main/plugin-health-scoring/Jenkinsfile#L1

alecharp:
Thanks @daniel-beck. Just improved the code to not fail if the report is not available.
I would like to take some time to see if we could remove the health score integer completely from the generated file if we didn't get a score. Cause, if we provide a 0 because reports service is down, that seems bad.

src/main/java/io/jenkins/update_center/HealthScores.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private static class JsonResponse {
+        public Map<String, PluginResponse> plugins;
+
+    }
+    private static class PluginResponse {
+        public int value;
+    }
+
+    public Integer getHealthScore(String pluginId) {
```
alecharp, line 59: I would prefer not to have the boxing / unboxing but the current fastjson configuration do not serialise `null` field. Using an `ObjectSerializer` with another _default_ value (like `-1`) is not working as only the value can be ignored from the serialisation, but the field name is generated anyway.

src/main/java/io/jenkins/update_center/HealthScores.java
```diff
@@ -0,0 +1,62 @@
...
+        return this.healthScores.getOrDefault(pluginId, null);
```
daniel-beck, line 60:
```suggestion
        return this.healthScores.get(pluginId);
```

does the same.

src/main/java/io/jenkins/update_center/HealthScores.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private static void initialize() {
+        final Request request = new Request.Builder().url(HEALTH_SCORES_URL).get().build();
+
+        try {
+            final String bodyString = HttpHelper.getResponseBody(new OkHttpClient(), request);
+            final JsonResponse response = JSON.parseObject(bodyString, JsonResponse.class);
+            if (response.plugins == null) {
+                throw new IllegalArgumentException("Specified popularity URL '" + HEALTH_SCORES_URL + "' does not contain a JSON object 'plugins'");
```
daniel-beck, line 40: This exception type bypasses the `catch` block, so unexpected JSON content will still break UC generation.
```suggestion
                throw new IOException("Specified popularity URL '" + HEALTH_SCORES_URL + "' does not contain a JSON object 'plugins'");
```

src/main/java/io/jenkins/update_center/HealthScores.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+import com.alibaba.fastjson.JSON;
+import io.jenkins.update_center.util.HttpHelper;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+
+public class HealthScores {
```
daniel-beck, line 15: Javadoc for this would be useful.

daniel-beck (commented):
Looks reasonable 😃

daniel-beck:
On-holding this, pending the corresponding core change.

src/main/java/io/jenkins/update_center/HealthScores.java
alecharp: of course. Thanks.

src/main/java/io/jenkins/update_center/HealthScores.java
alecharp: Correct. Thanks.

src/main/java/io/jenkins/update_center/HealthScores.java
alecharp: Just added what made sense to me. Please let me know if it's alright or if I need to improve it.

alecharp:
It now also remove OkHTTP dependency requirement.

alecharp:
@daniel-beck do you think we could move forward with this pull request, as the pull request on Jenkins seems to have reach an acceptable state.

daniel-beck:
So far I've been unable to have a successful execution, seems something's blocking/deadlocked somehow in the HTTP requests. Will need to look into this further later.

daniel-beck:
I switched generation of update sites to a snapshot uploaded from this branch, so currently, real update sites have `healthScore`.

src/main/java/io/jenkins/update_center/PluginUpdateCenterEntry.java
```diff
@@ -191,6 +191,10 @@ public int getPopularity() throws IOException {
         return Popularities.getInstance().getPopularity(artifactId);
     }
 
+    public Integer getHealthScore() {
```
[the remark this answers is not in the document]
dduportal: It would be a welcomed improvement BUT it is not important right now as we do not pay for download bandwidth today (Cloudflare does ;) ).

alecharp:
and I corrected the key in Jenkins Core pull request accordingly.

PR #841, opened 2025-02-03
Make `IssueTrackersSource` more resilient

daniel-beck:
https://github.com/jenkins-infra/update-center2/pull/840#issuecomment-2631201109 points out that the build fails if `reports.jenkins.io` cannot be reached. While the code was written to be more resilient, it wasn't done properly, this PR fixes that.


alecharp:
Not that it count but +1

MarkEWaite (approved):

dduportal (approved):
Thanks! Good catch \o/

PR #842, opened 2025-02-05
🚧 Modernize project configuration

alecharp:
This is about modernising the project configuration, upgrading the JVM used to compile and remove some unnecessary libraries in the process.

I wanted to change class defined for JSON parsing into records, but FastJSON doesn't seems to be working well with that. I'll try to migrate the project to FastJSON 2 or to Jackson 2 and see if that helps.

For now, I don't consider this complete.

Things I would like to work on here:

- [ ] Migrate classes to records
- [ ] Removes all spotbugs findings

### DIFF
--- a/src/main/java/io/jenkins/update_center/DirectoryTreeBuilder.java
+++ b/src/main/java/io/jenkins/update_center/DirectoryTreeBuilder.java
@@ -8,6 +8,7 @@ import org.kohsuke.args4j.Option;
 import javax.annotation.CheckForNull;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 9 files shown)


daniel-beck:
I'd tentatively suggest that OKHTTP removal and the future tasks be done in a subsequent PR to ease reviewing. You can stil build on top of the same commits as needed, but each subsequent PR's review surface will be smaller.

Additionally, a minimal "make this work for Java 21" could be merged early, to get CI for subsequent PRs, otherwise the Jenkinsfile won't be happy.

alecharp:
See #843 for the dependency removal.

PR #843, opened 2025-02-06
Removes usage of OkHTTP

alecharp:
This also eliminates Kotlin dependencies


alecharp:
This cannot be merge before #840 and it won't be working once #840 will be merged.

alecharp:
Yes if you want. Even if it's 2 different subjects.
We could merge the health score addition and see how it behaves for a will.
Then we can merge this and have the same observations.

But I cannot attend the pull requests in the next 2 weeks.

alecharp:
We can close this as it's now incorporated with https://github.com/jenkins-infra/update-center2/pull/840/commits/8c1381ec96e4c63b58a6f2a3202eb212857623be in #840.

PR #844, opened 2025-02-06
Update JVM from 11 to 21

alecharp:
This lets the project use Java 21. 
As the project deals a lot with String, this updates could help as the recent versions improved the memory management for that.

It opens us to bump support-log-formatter and version-number next (or let dependabot do it).

Once we have the build setup with this, we can go through the code and update API calls, removing deprecated onces, use. try-with-resources, records, and other few cleanups.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,1 @@
+--show-version
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,8 +14,8 @@ node('linux') {
     stage('Generate') {
         withEnv([
                 "PATH+MVN=${tool 'mvn'}/bin",
-                "JAVA_HOME=${tool 'jdk11'}",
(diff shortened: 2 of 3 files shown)


dduportal:
> This lets the project use Java 21. As the project deals a lot with String, this updates could help as the recent versions improved the memory management for that.
> 
> It opens us to bump support-log-formatter and version-number next (or let dependabot do it).
> 
> Once we have the build setup with this, we can go through the code and update API calls, removing deprecated onces, use. try-with-resources, records, and other few cleanups.

What would be the "threshold" in which to change the job configuration on trusted.ci.jenkins.io from JDK8 to JDK21?

daniel-beck:
> What would be the "threshold" in which to change the job configuration on trusted.ci.jenkins.io from JDK8 to JDK21?

Should be possible today. Otherwise only once the next actual release is out (or a snapshot build is used).

PR #847, opened 2025-02-18
Removing advisory for SECURITY-2774

stuartrowe:
Fix released in [jira-steps-plugin 2.0.180.vccfe35b_5910d](https://github.com/jenkinsci/jira-steps-plugin/releases/tag/2.0.180.vccfe35b_5910d) (see [jira-steps-plugin #190](https://github.com/jenkinsci/jira-steps-plugin/pull/190))

### DIFF
--- a/resources/warnings.json
+++ b/resources/warnings.json
@@ -14997,19 +14997,6 @@
     ]
   },
   {
-    "id": "SECURITY-2774",
-    "type": "plugin",
-    "name": "jira-steps",
-    "message": "Keys stored in plain text",
-    "url": "https://www.jenkins.io/security/advisory/2023-01-24/#SECURITY-2774",
-    "versions": [
-      {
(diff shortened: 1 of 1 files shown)


daniel-beck:
Thanks for letting us know! As this change would remove the security warning even for past, affected releases, we cannot merge this as is. I'll keep this open as a reminder to file a PR updating the regex. We have a tool that generates the regex for us, so no need for you to do that unless you're bored.

FWIW, interesting approach, I'll need to bookmark that and see what this does in detail. In general all that we expect is a field type change from `String` to `Secret` since that addresses the problem on the next save.

SNanda8895:
I reviewed the fix, and it correctly masks the API key.

daniel-beck:
Closing in favor of https://github.com/jenkins-infra/update-center2/pull/852.

PR #853, opened 2025-03-18
Exclude audit-trail version 393.v6a_3d1e251274 from distribution

PierreBtz:
It prevents instances from starting up (JENKINS-75425). Fixed by https://github.com/jenkinsci/audit-trail-plugin/pull/170


MarkEWaite (approved):
Thanks!

daniel-beck:
Seems reasonable to merge, although I would not generally remove historical releases from distribution.

daniel-beck:
Does this only apply to the single release being suspended though? What happens if I manually download and install 382 on 2.492.1?

PR #681, opened 2023-02-02
Deprecate `ace-editor`

daniel-beck:
As of https://github.com/jenkinsci/workflow-cps-plugin/releases/tag/3611.v201b_d9f9eb_f7, Pipeline: Groovy Plugin no longer depends on `ace-editor`, making it a library plugin without dependents that can safely be uninstalled by everyone who's keeping their Jenkins up to date.

This is the last plugin in https://github.com/jenkinsci/js-libs/ to be deprecated (after #679 and #677); perhaps we can consider archiving this repo?

Maintainers @batmat @rsandell to approve.

Thanks @basil for making this possible.

PR #857, opened 2025-03-29
suspend javascript GUI libs

mawinter69:
These plugins are not used by any other plugin. So to avoid that you even get them in the Available plugins the distribution should be stopped.
The repository is also archived.

Amends https://github.com/jenkins-infra/update-center2/pull/681 and https://github.com/jenkins-infra/update-center2/pull/677


mawinter69:
Not added the `jqeury-detached`plugin as [valgrind](https://plugins.jenkins.io/valgrind/) plugin is reported as depending on it.

MarkEWaite:
I thought that we didn't suspend distribution just because a plugin is archived or deprecated.  The ace-editor plugin (as one example) is currently installed on 115 000 controllers with over 5000 installs on 2.462.3 alone.  If we stop distribution and those users are installing with configuration as code, won't their configuration be broken?

daniel-beck:
Deprecation dates to >2 yrs ago. I see nothing wrong with removing these plugins at this point.

> currently

As of ~5 months ago, and that's only two thirds of what it was a year before.

> installing with configuration as code, won't their configuration be broken

Breaking configurations even when e.g. a field has been removed from configuration is a deliberate design decision by the CasC developers. As these plugins no longer serve a purpose, this basically reveals a bug in their CasC configurations.

MarkEWaite (approved):

daniel-beck:
FWIW I think we can also consider suspending `jquery-detached` and its dependents.

PR #858, opened 2025-04-12
Update dependencies

daniel-beck:
Since this tool now uses Java 21 since https://github.com/jenkins-infra/update-center2/commit/d117fa4f9d28639a2ff7d2cc546f7d6fa11df943 as a prerequisite to dependency removal in https://github.com/jenkins-infra/update-center2/pull/840, we can update some more dependencies, plus additional regular updates.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
-    <version>1.128</version>
+    <version>1.131</version>
   </parent>
 
   <artifactId>update-center2</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -164,14 +154,12 @@
     <dependency>
       <groupId>io.jenkins.lib</groupId>
       <artifactId>support-log-formatter</artifactId>
-      <!-- 1.3 requires Java 17 -->
-      <version>1.2</version>
+      <version>1.4.1</version>
```
MarkEWaite, line 157: I don't see a 1.4.1 release of support log formatter in its [releases](https://github.com/jenkinsci/lib-support-log-formatter/releases).  There is a 1.3 release
```suggestion
      <version>1.3</version>
```

MarkEWaite (commented):
I was worried that the ci.jenkins.io build failure was a problem specific to it, but I can duplicate the same failure on my local compilation.  When I switch to support-log-formatter 1.3, the compilation passes and tests pass.

pom.xml
daniel-beck: Thanks, looks like I screwed up my local Maven repo somehow.

MarkEWaite (approved):

PR #862, opened 2025-04-14
Follow HTTP redirects

daniel-beck:
By default, redirects are not followed, and HTTP 302 is considered a failure to download a few lines down.

This looks like it would explain the recent problems with cached download failures.

Amends https://github.com/jenkins-infra/update-center2/pull/840

CC @alecharp


alecharp:
I'm really sorry I created so many issues.

PR #868, opened 2025-05-24
Deprecate JaCoCo and Cobertura

zbynek:
https://plugins.jenkins.io/cobertura/
https://plugins.jenkins.io/jacoco/

All functionality of these plugins was migrated into https://plugins.jenkins.io/coverage/ and their repositories have been archived.

In 2.511 Cobertura plugin won't work because of https://github.com/jenkinsci/jenkins/pull/10628.
JaCoCo was removed from the list of plugins that are tested to work with current Jenkins releases: https://github.com/jenkinsci/acceptance-test-harness/pull/2017

The purpose of this PR is to provide a warning for all current users of those plugins as they still both have over 10k installs.


daniel-beck:
This seems redundant, given the plugins are already marked deprecated via GH label, and their docs provide information on it.

daniel-beck:
There is:

> Screenshot 2025-05-24 at 22 17 41

> Screenshot 2025-05-24 at 22 16 17

The only difference in presentation between label-based deprecation (here or via GH) and the dedicated deprecations file with URL is a slight difference in wording on the plugin site, and where the admin monitor and banner link to (plugin site page vs. deprecation URL).

PR #877, opened 2025-07-09
SECURITY-3353 Remove warning for dingding notifications

BobDu:
fixed by https://github.com/jenkinsci/dingtalk-plugin/pull/342

released in https://github.com/jenkinsci/dingtalk-plugin/releases/tag/dingding-notifications-2.7.4


resources/warnings.json
```diff
@@ -18542,7 +18542,7 @@
     "versions": [
       {
         "lastVersion": "2.7.3",
-        "pattern": ".*"
+        "pattern": "2[.]([0-6][.][0-9]+|7[.][0-3])"
```
daniel-beck, line 18545:
```suggestion
        "pattern": "(1|2[.][0-6]|2[.]7[.][0-3])(|[.-].+)"
```

* Adds 1.x
* Minimizes chars for older versions
* General suffix matches existing patterns in this file

Only the 1.x support is really needed though AFAICT.

daniel-beck (changes requested):
Thanks for letting us know!

resources/warnings.json
daniel-beck: groupIds can change, only artifactId matters for plugin identity. https://updates.jenkins.io/download/plugins/dingding-notifications/ lists also 1.1, 1.4, and 1.9 from https://repo.jenkins-ci.org/artifactory/releases/com/ztbsuper/dingding-notifications/

daniel-beck (approved):

PR #880, opened 2025-08-01
suspend commons-lang

mawinter69:
commons-lang is EOL since 2011 and vulnerable (CVE-2025-48924). The library plugin is not used by any other plugin. Due to ongoing activities to remove commons-lang from core and all plugins it will also not be needed via an implied dependency.


jtnord:
> That will not actually accomplish much then, because almost everyone will have it installed for all time.

It allows incremental delivery of the removal from core.  (a detached plugin does not need to be bundled IIRC, and it could then be up to the user to manually install it if they have an issue with a plugin that has not migrated).

If you are worried about it existing for all time, then once all known plugins have been migrated the plugin could self destruct (ie uninstall itself, show an admin monitor etc etc)

PR #881, opened 2025-08-05
feat(publish) stop publishing on the 'unsecured' (HTTP) web services as HTTPS is enforced

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4758


dduportal (commented):
Self merging to perform the operation

dduportal:
> Have you considered opening PRs from origin branches and switching the job over to your branch for changes only in the wrapper? That way you can test your changes without cluttering up history so much.

Did not though about this, but that would be a really good improvement, sorry for the "clutter". I'll use branch next time!

PR #883, opened 2025-08-07
chore(Jenkinsfile) use safer node label on ci.jenkins.io

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4762#issuecomment-3163158212


daniel-beck:
Is maven-21 really the label? Not jdk-21?

PR #888, opened 2025-09-20
prqa-plugin deprecated and replaced by https://plugins.jenkins.io/p4sa

isanych:


resources/artifact-ignores.properties
```diff
@@ -960,3 +960,6 @@ crossbrowsertesting = https://github.com/jenkins-infra/update-center2/pull/870
 
 # unused, EOL and CVE-2025-48924 
 commons-lang-api = https://github.com/jenkins-infra/update-center2/pull/880
+
+# replaced by https://plugins.jenkins.io/p4sa
+prqa-plugin = https://raw.githubusercontent.com/jenkinsci/prqa-plugin/refs/heads/master/README.MD
```
daniel-beck, line 965: Why the raw file content? We show a link there to users, not machines.

```suggestion
prqa-plugin = https://github.com/jenkinsci/prqa-plugin/blob/master/README.MD
```

PR #912, opened 2026-04-07
Update Center: mark docker-swarm-plugin as deprecated

rantoniuk:
Following https://github.com/jenkinsci/docker-swarm-plugin/pull/141.
I'm not sure if this should go in `label-definitions.properties` (as in the docs) or in the [deprecations.properties](https://github.com/rantoniuk/jenkinsci-update-center2/blob/master/resources/deprecations.properties)

/cc @roemer


daniel-beck:
I added the `deprecated` label to the GH repo, which is preferable over metadata overrides in this repo, when possible.

PR #924, opened 2026-07-29
Suspend signpath@5.0.0 due to accidental release with user disruption

Taha-cmd:
Version `5.0.0` of the signpath plugin was accidentally released with unfinished changes that caused disruption for users. Please remove it from the update center.


daniel-beck:
I'd like to know more than that to suspend a release. There are no open issues or open PRs that explain why this needs to disappear.

In general, we recommend you fix forward, i.e., fix whatever is wrong and create a new release with the fixes. Suspending releases should be a last resort, not the first thing you do. This will also ensure users who have already upgraded will be taken care of.

Taha-cmd:
Version 5.0 was not ready for release yet. It's not usable as it is right now and fixing forward is not possible in a timely manner due to outstanding architectural decisions and other dependencies. The fix will most likely also contain breaking changes, which basically means another major version bump. We already got support requests on our support email. Currently, users either have to pin version 4 (which is not possible to do with the gallery; they'll have to manually upload the HPI file) or wait for version 6 to be released. This makes removing version 5 from the gallery the best option for us.

MarkEWaite:
> Version 5.0 was not ready for release yet. It's not usable as it is right now and fixing forward is not possible in a timely manner due to outstanding architectural decisions and other dependencies.

One way to "fix it forward" is to create a 5.1 release that uses the same source code as the most recent working 4.x version.  That allows users to upgrade to 5.1, allows you to continue development and testing at a pace that works for you, and does not require that users who have already installed 5.0 must perform complicated operations to remove 5.0.

paulsavoie:
Hello @MarkEWaite, jumping in here. 

The reason why we don't want to publish a 5.1 release is that we follow semantic versioning. Unfortunately, with Jenkins plugins, there seems to be no update warning if you upgrade to a higher major version, that's why we are worried that customers will just update without checking (just click "Update" in Jenkins). This already happened to one customer who downgraded manually.

The current 5.0 release that was published is indeed a breaking change for customers (that's why we increased the major version number). If we now release a 5.1 reverting the old behavior, that would be wrong from a semver perspective (which we value very much).

Is there any way we can remove the release instead of having to mess around with our versioning like that? I also don't want to release a 6.0 version that has the same interface as the 4.0 version, as this would be confusing for our customers.

MarkEWaite:
> The current 5.0 release that was published is indeed a breaking change for customers (that's why we increased the major version number). If we now release a 5.1 reverting the old behavior, that would be wrong from a semver perspective (which we value very much).

I think that you are placing more value on semantic versioning than on the experience of your users.  If you suspend distribution of the 5.0 release, that prevents new installations from receiving 5.0 but it does not solve the problem for users that installed the 5.0 release already.  Those users must perform a series of steps to return to the previous version.  They may need to perform additional steps in the future, depending on how you number future versions.

If necessary for your desire to use a more pure form of semantic versioning, call the new release 6.0 when you revert to the previous behavior.  It is incompatible and a breaking change from 5.0, so name it 6.0.  That allows users a simple path to the new version without extra steps to remove the 5.0 version.

zbynek:
> Unfortunately, with Jenkins plugins, there seems to be no update warning if you upgrade to a higher major version,

Btw. the next time you're releasing a breaking change you can mark a version as incompatible with older releases:
https://www.jenkins.io/doc/developer/plugin-development/mark-a-plugin-incompatible/

